PR #1, opened 2023-06-20
ADR for architecture guidance

jenniferplusplus:
I hope you're familiar with ADRs. If not, the basic idea is to write down both the more significant choices made in a project, as well as the reasoning that went into them. And they often go in source control because as programmers we are obligated to use that particular hammer for every single nail.

So with this, I guess I want to know if the logic and the choices make sense and seem reasonable, or if you can think of anything I missed.

### DIFF
--- a/docs/decisions/00-template.md
+++ b/docs/decisions/00-template.md
@@ -1,0 +1,24 @@
+﻿# Subject
+
+## Status
+
+- [ ] Decided (proposal was agreed and adopted)
+- [ ] Decided Against (agreed not to adopt)
+- [ ] Deferred (no decision, but should be revisited later)
+- [ ] Superseded (later events or decisions make this no longer relevant)
+
+# Decision
+
+A summary of what was proposed and decided
+
+## Impact
+
+Details about what changes because of the decision
+
+## Context
+
+Information about the situation that prompted the proposal/decision
+
+## Discussion
+
+A summary of the comments/issues/concerns that led up to the decision
--- a/docs/decisions/01-system-design-guidance.md
+++ b/docs/decisions/01-system-design-guidance.md
@@ -1,0 +1,48 @@
+﻿# System Design Guidance
+
+Priorities for the project, and some guidance on how to balance those priorities when making high level architectural decisions.
+
+## Status
+
+- [ ] Decided (proposal was agreed and adopted)
+- [ ] Decided Against (agreed not to adopt)
+- [ ] Deferred (no decision, but should be revisited later)
+- [ ] Superseded (later events or decisions make this no longer relevant)
+
+# Decision
+
+This is about guidance, so here's the guidance.
+
+1. Prefer multiple providers. There will be system components that can be rented from managed service providers (you know, clouds). It's better to choose technologies that are available in multiple cloud providers. In the future, it would be even better to support multiple services in a category (MySQL AND PostreSQL, for example). That's not feasible now, but try not to design out the option of doing it later. 
+2. Be self-hostable. Don't choose services that are only available in the cloud (like snowflake or dynamodb), or that are difficult to administer (like kafka). This goes double if it's only available from a single provider.
+3. Plan for scale-up. Networks tend to grow over time. It should be a reasonable amount of work for a volunteer admin to add capacity as the need arises.
+
+
+## Impact
+
+Now
+* DB - Postgres
+* Object Storage - Any S3 compatible
+
+Future
+* Message Queue - Something internal (maybe subscriber pattern?), then RabbitMQ (if/when we need it)
+* Orchestration - Something internal (threads, processes), then maybe Nomad?
+
+## Context
+
+Basically, this came up when I was considering which database engine should serve as the core database for the project. There are a number of goals for Letterbook, and I want to write them down, for the future.
+
+### Optimize for the operators
+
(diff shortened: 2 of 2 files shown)


docs/decisions/00-template.md
runewake2, line 1: When cataloging these decisions through git, how do you usually capture historic conversation about a feature (I've typically done this process through a Google/Word Doc)? Do we care about that or is this what the `Discussion` section is for?

docs/decisions/01-system-design-guidance.md
```diff
@@ -0,0 +1,48 @@
...
+- [ ] Deferred (no decision, but should be revisited later)
+- [ ] Superseded (later events or decisions make this no longer relevant)
+
+# Decision
+
+This is about guidance, so here's the guidance.
+
+1. Prefer multiple providers. There will be system components that can be rented from managed service providers (you know, clouds). It's better to choose technologies that are available in multiple cloud providers. In the future, it would be even better to support multiple services in a category (MySQL AND PostreSQL, for example). That's not feasible now, but try not to design out the option of doing it later. 
+2. Be self-hostable. Don't choose services that are only available in the cloud (like snowflake or dynamodb), or that are difficult to administer (like kafka). This goes double if it's only available from a single provider.
+3. Plan for scale-up. Networks tend to grow over time. It should be a reasonable amount of work for a volunteer admin to add capacity as the need arises.
```
runewake2, line 18: Will this mean that our servers need to support horizontally scaling (potentially with an exception for the DB) from the start?

docs/decisions/01-system-design-guidance.md
```diff
@@ -0,0 +1,48 @@
...
+
+## Context
+
+Basically, this came up when I was considering which database engine should serve as the core database for the project. There are a number of goals for Letterbook, and I want to write them down, for the future.
+
+### Optimize for the operators
+
+There is some some threshold of features that would make Letterbook viable for users. Beyond that, we should favor the needs of instance administrators, operations teams, and moderators. That group is likely to overwhelmingly be hobbyists and volunteers. They will often have shoestring budgets. They will probably live all over the world, speak numerous languages, and work in various time zones. In particular, I'm thinking of the experience that the https://hachyderm.io team had in setting up and scaling up their mastodon instance. They had a team of frankly world class SREs, and they still struggled with it continuously for weeks. Many instances had to close registrations during the twitter migration because they couldn't operationally support the added demand. You shouldn't have to have a team of senior engineers to run a community chat service.
```
runewake2, line 37: How do you see this impacting Telemetry and Observability? Adding that could make maintenance and upkeep simpler, *but* can lead to some pretty substantial costs for storing and cataloging all of that telemetry info which might conflict with the next goal.

Will we need a system for being able to toggle features on/off easily that Admin's can use?

docs/decisions/01-system-design-guidance.md
```diff
@@ -0,0 +1,48 @@
...
+
+### Minimize cost
+
+Letterbook should be maximally affordable, because people use these services for community projects, and fund them on donations. Businesses can just throw RAM or IOPS at problems until they go away, but that's not likely to be an option here. So system components need to actually be suited to the task and scale.
+
+
+## Discussion
+
+So, I think those technology choices are pretty reasonable. Except maybe Postgres. I don't *love* using Postgres for this. I considered a document db. I think the ones that could be viable are Mongo or CouchDB. The problem with mongo is that I can't find a managed provider that I think has a reasonable price point. And the problem with CouchDB is that the only managed provider I can find is IBM. I also considered a graph database. It's a social network, after all. The network graph is the key thing here. The problem again is that the only reasonable option is Neo4j, and it's super expensive to get as a managed service.
```
runewake2, line 46: I don't mind Postgres. If we need some sort of blob storage they support [json](https://www.postgresql.org/docs/current/datatype-json.html) and I think that should be fine. I'd expect scaling wise this isn't where we'd hit limits.

docs/decisions/00-template.md
jenniferplusplus: That's what I was imagining the discussion section would do. We can always link back to PRs and issues, too. At work it might go in a google doc, but with an open source hobby project, it's hard to be sure that would continue to be available in the long term.

docs/decisions/01-system-design-guidance.md
jenniferplusplus: I think we should expect to have to scale horizontally, and design to facilitate that. But I think a design that's amenable to horizontal scaling in the future is all we would need for now.

docs/decisions/01-system-design-guidance.md
jenniferplusplus: Yes, letterbook should produce some telemetry. I'm a fan of OpenTelemetry. When we get to that point, I'm imagining a guide or template or something that includes an OTel collector, with configs to make it easy to adjust sampling rates to keep that cost down.

jenniferplusplus: Oh, for feature flags/feature toggles/something like that: maybe? I think there's a lot of value in those systems, but I normally do those things in the context of a service I'm operating. I'm not sure how I would go about it in this case.

runewake2: One thing I'm wondering here is how much can we defer to plugins for feature support. If the core of this system is just an MVP with basic functionality and everything else is left up to plugins, this cost and optimization knob might be easier to turn (thinking sort of the WordPress model here). I'm not familiar enough yet to know exactly what that would look like, but I could see it enabling a lot of different useful things and allow admins to customize their servers to their needs. That approach might encompass the idea of feature flags to and make them redundant. This might come with some up-front complexity in setup, but it'd mean you could choose from say an elastic powered search feature or just an indexed hashtag powered one instead of us making that decision for you.

jenniferplusplus: That's an interesting idea. I've seen plugin driven systems go very badly before. Or, once. But clearly they can work. I think that's probably a question of how to support contributors. If this takes off like a rocket and there's tons of people looking to build things, then plug-ins look really appealing. If it's more gradual, or people just want to use letterbook, not build it, then my thinking is they're more trouble than they're worth.

runewake2: Maybe it's an idea worth revisiting after an MVP. I imagine we'll learn a lot about what works and doesn't work building that.

runewake2 (approved):

PR #11, opened 2023-06-21
Put Snake_Case Converter under test and add initial CI workflows

runewake2:
Adding an xunit project to test the project and place the snake_case converter under test.

Add two CI jobs to validate pull requests against `main` and pushes made to `main`


jenniferplusplus (approved):
Love it

PR #13, opened 2023-06-22
Rename CI job

jenniferplusplus:
Github isn't letting me add it as a branch protection requirement, and I'm hoping it's because of the spaces in the name


jenniferplusplus:
I needed to specify the job, not the workflow. So all good already

Issue #3, opened 2023-06-21
Initial ActivityPub data models


jenniferplusplus:
- [x] Models to represent AP actors, objects, and activities (at least enough for a single simple activity)
- [x] Serialize to JSON-LD (at least)
- [x] Deserialize from JSON-LD (at least)

PR #15, opened 2023-06-24
Mock Actor implementation

jenniferplusplus:
This implements the GET /actor/{id} endpoint with a mock response. It's not particularly useful as-is, but I wanted to figure out a good way to build those collection URLs.

This also includes a mock implementation for POST /actor/shared_inbox. It doesn't do anything other than log part of the request, but I think that should be enough to start receiving federated public posts if we can get a stream of them.

## Related
- #3


jenniferplusplus:
@runewake2 I'm going to merge this, but tagging you to make sure you see this so you can stay up to date

runewake2:
Thanks, lgtm!

PR #29, opened 2023-07-05
Initial CoC

### LINKED ISSUE #6: CoC

jenniferplusplus:
I particularly want to discuss this one, since it's the code of conduct. I think CoC's really exist in their enforcement, and doing a copy & paste of the generic contributor covenant or similar doesn't do much to convey to people how we would approach that. So, I took a different approach.

Please let me know if you think anything here is too much, or off base, or if something is missing.

# Related
- closes #6


runewake2 (approved):
Signed. LGTM.

jenniferplusplus:
Cool. Did you want to include anything else, or have any issue at all? I'm trying _not_ to dictate things.

runewake2:
It seems in line with other CoC's like the [Contributor Covenant](https://www.contributor-covenant.org/) and I think the wording is clear.

I think it already covers the main thing: we're stronger as a community than individuals and should work in a way that promotes that cooperation and inclusion. imo this outlines that well already. Only potential change is making the "who do I contact if things go poorly" more discoverable so those that need it can find it.

As this grows we might want to make someone own this in the case where a maintainer crosses a line they shouldn't. But I don't think that's an immediate need right now.

jenniferplusplus:
I was hoping that all of the core maintainers would be able to respond to a report, and we would confirm they're good with that before bringing them on. But yeah, we can revisit in the future, when that's not just hypothetical.

Issue #36, opened 2023-07-05
API Authorization


jenniferplusplus:
We need an authorization framework. A lot of authz can be handled at the api layer. We have a JWT based authentication mechanism already. We would preferably use that to implement OIDC. That's not trivial, but openiddict is a relatively known quantity. From there we can implement a claims based framework. 

But wait there's more!

First, we don't have a set of claims defined for ourselves, much less any kind of hierarchy or organization. We need that, but it can probably wait a little while.

Second, I had been thinking we need to also support mastodon style claims, but maybe we don't? Apps will most likely request them, but we can provide whatever claim we want in response to that. It would be good to confirm either way.

Third, we definitely need to support http signature headers as a combined server to server authn/authz mechanism. It is what it is.

So! To do:

- [x] setup policy based authz, or any other scheme that makes it easy to mix and match our methods.
- [x] share whatever you learn about claims and any thing else that's relevant
- [x] #223
- [x] HttpSignatures authentication scheme
[This article may help](https://matteosonoio.it/aspnet-core-authentication-schemes/)
[Or the aspnet security guide](https://learn.microsoft.com/en-us/aspnet/core/security/authentication/?view=aspnetcore-8.0)
- [x] ~~implement oidc with openiddict~~ (punting to the future, this is actually a major effort)

rytmis:
I'm interested in participating in this project, and this seems like something I might be able to help with. 

I don't know OpenIddict aside from what I've read from their docs, but I've shipped a number of IdentityServer deployments and worked with various oidc implementations. Not an expert by any means, but not completely lost either. And I at least think I have some sort of handle on the aspnetcore authn parts.

jenniferplusplus:
@rytmis that's great, I'd love to have some help with this! Of course, it's a big topic. I don't want to just throw you into the deep end with it. I've been giving authz some thought at a high level lately, and lot of that is captured in the[ draft ADR doc that's open for review](https://github.com/Letterbook/Letterbook/pull/183). Definitely have a look at that, I'd like to know your thoughts.

My sense is that OIDC is not urgent, but I also want to minimize how much trouble it is to integrate later. So I would also really like to have some thoughts on whether than can be delayed or not.

jenniferplusplus:
[...]

PR #37, opened 2023-07-06
ADR for accounts and profiles

jenniferplusplus:


docs/decisions/03-accounts-profiles-actors.md
```diff
@@ -0,0 +1,76 @@
...
+
+In the longer term, this also creates the possibility that accounts can be connected to more than one profile, and that profiles can be connected to more than one account. This can enable things like easier and more robust management of bot profiles, shared access to profiles representing organizations, or delegating moderation powers between trusted contacts.
+
+## Context
+
+This permits us to separate how people log in from how they would send and receive messages. That's nice for individuals, because it gives them more flexibility to divide up their online persona, if they want to. For instance, having personal and professional profiles. The other thing this enables is to allow a single profile to be controlled by multiple accounts. And I think that is critical to organizations. And that really applies to all organizations.
```
runewake2, line 70: I'm a really big fan of this separation and I think it enables some really fun future capabilities around OAuth and SSO that I haven't seen elsewhere. Potentially we should consider how multi-account profiles will have responsibilities managed and also consider access restrictions at a per-account level: for example staff that can read and respond privately, but not make public posts, or read-only, etc. Or for example giving a Bot permission to read and post for me, but not manage the profiles settings, delete posts etc.

Also agree with where you are separating these models. Profiles and Accounts is a good slice for this.

runewake2 (approved):
LGTM, 1 suggestion expanding the scope of this problem a bit around permissions. I think you're already getting at this, but might be handy to call it out explicitly.

docs/decisions/03-accounts-profiles-actors.md
```diff
@@ -0,0 +1,76 @@
...
+
+## Discussion
+
+A summary of the comments/issues/concerns that led up to the decision
```
jenniferplusplus, line 76:
```suggestion
This ADR largely ignores permissions and access controls surrounding shared profiles. That's an important and complex topic, and one that warrants careful consideration. In addition to this account/profile domain modelling, that will also depend on how we implement an authorization model.

[See the PR](https://github.com/Letterbook/Letterbook/pull/37/files)
```

@runewake2 reasonable?

runewake2: ✅ Agree, we can revisit when looking into #36

docs/decisions/03-accounts-profiles-actors.md
```diff
@@ -0,0 +1,78 @@
+# Accounts vs Profiles vs Actors
+
+Clarifying the difference between these concepts, and the value of having them be different.
+
+## Status
+
+- [ ] Decided
```
jenniferplusplus, line 7:
```suggestion
- [x] Decided (2023-07-10)
```

Issue #38, opened 2023-07-07
ActivityPub lib


jenniferplusplus:
Fedodo is not going to work long term, and it's kind of painful to use. I'm running out of ways to work around that, so I think I have to stop and handle the AP spec.

My basic plan is to parse them through JSON-LD.net, and use the "frame" transform to map them to a consistently deserializable shape. This would then preferably be made available as a set of [CustomConverters](https://learn.microsoft.com/en-us/dotnet/standard/serialization/system-text-json/converters-how-to?pivots=dotnet-7-0) for ASP.net

In this design, the bulk of the work probably lives in building good framing documents, which seems like the least of all evils in terms of dealing with the AP spec in statically typed languages.

- [x] Model types support JSON fields that can be missing/null/single valued/array valued
- [x] Model types support JSON fields that can be either Link or Object valued
- [x] Both at the same time
- [x] Deserialize to the correct types
- [x] Easily retrieve concrete types
- [x] Serialize single item lists as single values, and exclude null values
- [x] Read and process `@context` field
- [x] Serialize `@context` field correctly
- [x] Handle compact Iri values
- [ ] Support extensions to the spec
- [x] Tests
- [x] Create and publish nuget package somewhere

jenniferplusplus:
I made a _lot_ of progress on this. It turns out that JSON-LD framing documents are even harder to write than polymorphic json converters, and it would have garbage performance because general purpose JSON-LD processing requires at least 2 passes over the fully loaded document. So, I just did the converters.

jenniferplusplus:
Thinking about extensions. Maybe as part of that support, this `if` chain (and similar) can be replaced by iterating over a list of some kind of matcher objects, and that list can be built via reflection to get all the implementations in the loaded assemblies.

https://github.com/Letterbook/Letterbook.ActivityPub/blob/main/Letterbook.ActivityPub/ConvertResolvable.cs#L32-L55

jenniferplusplus:
This lib is built and published. Additional work is being tracked in that repo.

PR #41, opened 2023-07-14
Migrate to Letterbook ActivityPub DTOs

jenniferplusplus:
The Letterbook AP models will need a little bit of work before they're ready for prime time, but that's okay, the other lib needed more. They'll be *much* easier to work with, and I don't want to write a bunch of type mappers twice, so I'm doing this now.

### Related
- #38

### DIFF
--- a/Letterbook.Api/Controllers/ActivityPub/ActorController.cs
+++ b/Letterbook.Api/Controllers/ActivityPub/ActorController.cs
@@ -3,7 +3,9 @@ using Fedodo.NuGet.ActivityPub.Model.ActorTypes;
 using Fedodo.NuGet.ActivityPub.Model.ActorTypes.SubTypes;
 using Fedodo.NuGet.ActivityPub.Model.CoreTypes;
 using Letterbook.Core;
+using Models = Letterbook.ActivityPub.Models;
 using Microsoft.AspNetCore.Mvc;
+using Microsoft.EntityFrameworkCore.Infrastructure;
 using Microsoft.Extensions.Options;
 
 namespace Letterbook.Api.Controllers.ActivityPub;
@@ -91,7 +93,7 @@ public class ActorController
     [HttpPost]
     [Route("[action]")]
     [ProducesResponseType(StatusCodes.Status202Accepted)]
-    public async Task<ActionResult> SharedInbox(Activity activity)
+    public async Task<ActionResult> SharedInbox(Models.Activity activity)
     {
         // TODO: add dependency on ActivityService and call it here
         await _activityService.Receive(activity);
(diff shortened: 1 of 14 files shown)


Letterbook.Api/Controllers/ActivityPub/ActorController.cs
```diff
@@ -1,8 +1,8 @@
 ﻿using System.Reflection;
 using Fedodo.NuGet.ActivityPub.Model.ActorTypes;
 using Fedodo.NuGet.ActivityPub.Model.ActorTypes.SubTypes;
-using Fedodo.NuGet.ActivityPub.Model.CoreTypes;
 using Letterbook.Core;
+using DTO = Letterbook.ActivityPub.Models;
```
runewake2, line 5: Does it make sense to put some of this in the Global Usings, some of these models and Activity Pub related content feels like it will be commonly used and establishing a shorthand across all files in the project instead of hoping to copy it between each might help in the long term. Especially with the project already having duplicate names for things like `Actor` - it might cut down on confusion that could cause.

Letterbook.Core/Models/Actor.cs
```diff
@@ -9,9 +9,16 @@
 /// </summary>
 public class Actor
 {
+    private Actor()
+    {
```
runewake2, line 13: I'm not sure I understand mixing where these types are initialized (constructor vs property initializer). Can you help me understand this?

Letterbook.Core/Letterbook.Core.csproj
```diff
@@ -8,6 +8,7 @@
 
     <ItemGroup>
       <PackageReference Include="Fedodo.ActivityPub.Model" Version="2.0.27" />
+      <PackageReference Include="Letterbook.ActivityPub" Version="1.0.0-ci0010-39" />
```
runewake2, line 11: Is there any reason we wouldn't want this to stay in sync with new [Letterbook.ActivityPub](https://github.com/Letterbook/Letterbook.ActivityPub) packages? Would it be worth investing in automating that to prevent drift between these projects?

runewake2 (approved):
A few comments/questions on this one.

Letterbook.Api/Controllers/ActivityPub/ActorController.cs
jenniferplusplus: That's a good idea. I forget that global using exists, tbh.

Letterbook.Core/Models/Actor.cs
jenniferplusplus: This is an efcore thing. It does some special magic to get access to the private default constructor, uses that to create the object, and then uses the accessors to set values on everything.

So, this does some null forgiveness for the non-nullable fields, to satisfy the compiler. But if we're creating these objects in user code, we'll still get compiler errors about leaving them unitialized.

Letterbook.Core/Letterbook.Core.csproj
jenniferplusplus: For now, yes they should stay in sync.

runewake2 (approved):
LGTM, the global thing can either be added to this or we can do that later if we like.

PR #63, opened 2023-08-10
Clarify self determination

jenniferplusplus:
Clarify what self determination means, based on some discussion with other contributors


warriordog (approved):
Looks good to me
(disclosure - I co-wrote this)

PR #86, opened 2023-08-20
Implement Timeline service

### LINKED ISSUE #51: Timeline Service
- [x] Domain models
- [x] Add posts to audience feeds
- [x] Query logic (most of this is a concern for the adapter)
  - [ ] Dedupe
  - [ ] Paging
  - [ ] Anything else that seems obviously necessary
- [x] Support Actor inbox/outbox (mostly outbox)

jenniferplusplus:
This is the initial implementation of the core timeline service.

## Details
- See the updated timeline.md doc
- This part isn't implemented yet, but the Timeline service will play a large role in the enforcement of moderation actions. Blocks and mutes and so on will impact whether any given post is added to a feed (ideal) and whether it is returned from a query

## Related
- Closes #51


Letterbook.Core/Models/MentionVisibility.cs
```diff
@@ -6,4 +6,5 @@ public enum MentionVisibility
     Bcc,
     To,
     Cc,
+    // does this also need to include Audience and Tags?
```
runewake2, line 9: Maybe a good idea to spin this off into an issue and tag it here so it isn't lost and we can keep visibility on any discussion.

docs/timeline.md
```diff
@@ -34,9 +31,35 @@ erDiagram
...
+
+Instead, we'll be using a timeseries database. Specifically, TimescaleDB, at least to start. Timeseries DBs are analytic—as opposed to transactional—databases. The basic premise is that by giving up on relational integrity, the database can very quickly select, filter, and aggregate very large sets of records. And timeseries DBs in particular, are designed around doing this with data that is ordered and queried chronologically. Exactly like a timeline.
```
runewake2, line 44: What are the operational impacts of this for admins? I really like this choice, but people may be more familiar with operating relational SQL databases vs a TSDB.

docs/timeline.md
```diff
@@ -34,9 +31,35 @@ erDiagram
...
+
+## Details
+
+But, again, there's no relational integrity with any of this. So looking up a post's creator's followers to determine whether it should be excluded by this WHERE is a non-starter. This means we need to do some pre-processing of our data to make it easy to query without those relations. The solution to that is the `Audience` concept. Before they're added to the tsdb, posts get lightly pre-processed to generate audience keys, and these keys essentially constitute feeds, which can be easily merged at query time. Typical audiences would be `public`, the creator's `followers`, and the `mentions` for anyone mentioned in the post. Then to populate a feed, the timeline service will query the feeds table with something like this:
```
runewake2, line 48: It seems like this is going to create a really high amount of cardinality in the system (a key per user?), am I understanding that correctly?

docs/timeline.md
```diff
@@ -34,9 +31,35 @@ erDiagram
...
+
+If you noticed that this means we'll be duplicating a lot of data in this table, you get a gold star. This is again not a significant concern thanks to the properties of timeseries databases. There are two factors behind this. One, is that TSDBs can achieve extremely good at-rest compression by using column-oriented storage. Two, these databases are very good at ejecting old data. In fact, if you're fine with letting the db do whatever is easy (and I am), this happens basically for free. Everything is already time-ordered and stored as columns on disk. Those columns are already further partitioned by time, as a query optimization. Because of this, partitions that are older than your chosen expiration time can just be discarded with practically no effort.
```
runewake2, line 65: I'm not sure I understand the implications here. If things are automatically removed from the DB what happens to a users personal feed over time? Will they lose access to data? Also, a big cost in cloud-ops can be attributed to storage size, especially for a larger instance this may not be a trivial cost.

docs/timeline.md
jenniferplusplus: Timescale is actually a plugin for Postgres, so the operational concern is installing that plugin. Or providing connection info for a managed instance. I plan to put feeds in its own database. It can be on the same postgres instance as everything else, or a different one. And it shouldn't be too difficult to migrate from a shared to dedicated instance when a server grows to that size.

docs/timeline.md
jenniferplusplus: At least one key per known profile that's posted something since the expiration window, yes. The whole fedi network only has 10m profiles, and only about 2m of them are active. Even if they all were active and followed by members on a single server that's only a cardinality of about 10 million. TBH, I'm not sure how timsecale handles very large cardinality, but I know Influx can handle well into the hundreds of millions, and influx v3 is supposed to be able to handle unlimited cardinality.

Which is to say I'm confident the pattern can scale as far as it will ever need to, even if timescale specifically can't.

Full disclosure, I work at Influx and would have just chosen that, but it fails on the multi-vendor guidance for the project, and doesn't have a great self-hosting story either. I also like that timescale lets small servers operate with just the service and a postgres database.

docs/timeline.md
jenniferplusplus: That's a good point about a single user's timeline. It could be regenerated from the regular relational store, but that might not be desirable. We might need to exempt those feeds from expiring, or find other handling for it.

As far as the storage used by the TSDB, that should stay small. The compression really is bonkers good, and we can certainly eject duplicates after some time. For example, public posts will generally go to the followers and public audiences. We can expire the public duplicate. And that expiration window is easy to configure.

https://www.timescale.com/products#controlCosts

docs/timeline.md
runewake2: Appreciate that, would love to learn more from you about Time Series in the future!

If I'm understanding things: When someone posts we add a new record keyed with their profile? Are posts going to be keyed by the timeline they are in (ex. a post gets replicated into each users timeline) or by the owner of who posts them?

When I follow a new account or unfollow an account it seems like that could trigger a lot of work if we are creating distinct timelines per user (potentially replicating an entire post history + boosts every follow/unfollow).

Letterbook.Core/Models/MentionVisibility.cs
jenniferplusplus: Maybe. It's mostly an activitypub question. Or rather, it's mostly a question of how everyone else has implemented activitypub. It wouldn't make sense to me for them to reference individual actors in those fields, but it also wouldn't make sense to me for mastodon to put the `as:public` special audience in the `To` or `Cc` fields, but they do.

Letterbook.Core.Tests/TimelineServiceTest.cs
```diff
@@ -0,0 +1,384 @@
...
+    public void Exists()
+    {
+        Assert.NotNull(_timeline);
+    }
+
+    [Trait("TimelineService", "HandleCreate")]
+    [Fact(DisplayName = "HandleCreate should add public posts to the public audience")]
+    public void AddToPublicOnCreate()
+    {
+        var note = _note.Generate();
```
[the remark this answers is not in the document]
jenniferplusplus: I don't generally mind duplication in test code. I'd rather have duplication than add confusion by trying to extract all the common elements of test scenarios. But yeah, this one can just move up to the constructor.

docs/timeline.md
jenniferplusplus: Keyed based on the creator(s). Specifically the key will be the value for `Profile.FollowersCollection.Id`. And then the followers for that creator will query for that key. So when you follow a new account, that just adds one more item to the WHERE clause to build your home timeline.

And FWIW, I do devops at influx, not like core database stuff. I've picked up some things being in the environment, but I'm not a timeseries expert.

runewake2: Is there a point where adding more WHERE clauses becomes problematic? Will this choice lead to a limitation on follower count for example or degrade performance for users with significant follower counts in a UX impacting way?

jenniferplusplus: I don't *think* so. More WHERE clauses could, maybe, depending on the query. But I wouldn't expect more items in the WHERE in collection to cause a problem. Again, I'm running into the limits of my database internals expertise here, but I would expect the query planner to turn that into a hash join, which DBs do all the time.

runewake2: Makes sense, I'm not a DBA either and haven't used that tech. Sorry for all the questions, I do really like this idea though! I can spin this question off and take a look if we want. It seems like this would be a helpful thing to test but I don't think it's productive to block this.

jenniferplusplus: No worries at all! Asking questions is exactly what review is for. You're right that some of this is stuff we'll just need to test. I think we're pretty close to being able to do that in a meaningful way.

Issue #93, opened 2023-09-11
Initial Profile Service


jenniferplusplus:
Now:
- [x] create profile
- [x] modify profile
- [x] lookup profile
- [x] search for profiles
- [x] follow someone
- [x] be followed by someone
- [x] remove follower
- [x] unfollow
- [x] receive follow request
- [x] add to feed (aka join audience)
- [x] remove from feed (leave audience)

Punting this to a future moderation ticket:
- [ ] accept new follower
- [ ] reject follow request
- [ ] send follow request
- [ ] send report
- [ ] block
- [ ] mute

Punting these to future feature tickets:
- [ ] subscribe (follow, but only see public posts)
- [ ] transfer in
- [ ] transfer out
- [ ] delete profile
- [ ] grant access to another account
- [ ] revoke access

jenniferplusplus:
Note to self, change single audience key to follow/subscribe switch plus list of additional audience keys
https://github.com/Letterbook/Letterbook/blob/1cfc87e36668f84463bef41b2dabf8a47a17909c/Letterbook.Core/ProfileService.cs#L209C87-L209C87

PR #94, opened 2023-09-17
Initial, partial implementation of Profile Service

jenniferplusplus:
This scaffolds out a lot of stuff related to profiles and the profile service. This isn't done, but I'm trying to be better about making PRs a reviewable size.

## Details
- Implement and test Create methods
- Move away from using b62 encoded UUID as internal ID type. Just use GUID instead.
  - b62 encoding should still be used in DTOs, because it's still better for URLs
- Early work on making Exceptions consistent and helpful
  - The pattern is more or less final. Eventually I want them to have multiple language support, and help links. But that's a long way off, and a long long way off, respectively.

## Related
- #93


jenniferplusplus:
Ooh, look at that CI keeping e honest

PR #109, opened 2023-10-31
Introduce Serilog logger

runewake2:
Adding structured logging to hopefully help make debugging/monitoring a bit more approachable. This is a first pass - will probably need more thought in order to support additional sinks depending on where this is hosted.

## Details
- Introduces a dependency on [`Serilog.AspNetCore`](https://github.com/serilog/serilog-aspnetcore)
- Logging is configured in `appsettings.json` under `Serilog`. [More about how to do that here](https://github.com/serilog/serilog-settings-configuration).
- Mostly mirrors default .NET logging behavior but with a bit more structure.

I'm most familiar with `Serilog` for structured logging in .NET so that's what I've used here, there may be other options worth pursuing I'm less familiar with.

### DIFF
--- a/Letterbook.Api/Letterbook.Api.csproj
+++ b/Letterbook.Api/Letterbook.Api.csproj
@@ -23,6 +23,7 @@
       <PrivateAssets>all</PrivateAssets>
       <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
     </PackageReference>
+    <PackageReference Include="Serilog.AspNetCore" Version="7.0.0" />
     <PackageReference Include="Swashbuckle.AspNetCore" Version="6.4.0" />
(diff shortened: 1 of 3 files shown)


jenniferplusplus (approved):
I know Serilog has more transport options than microsoft.extensions.logging, but otherwise I don't know what the benefit is. I'm also not sure if we need those transports, tbh? But there's no real harm in having it (modulo something like log4shell).

I'm fine with this, but what I would ideally want is an observability stack we can run in development. So: log/trace/metric collection, and backends to query them (if it was me, I'd put together a grafana/loki/prometheus/tempo docker compose spec). That lets telemetry become part of the dev feedback loop, which is what I really want.

edit: we obviously don't have metrics or traces yet, so just logs via grafana/loki, or something else, would be a great start.

PR #112, opened 2023-11-03
Cleanup warnings and tests

jenniferplusplus:
Some general code maintenance. I want to eliminate all of the compiler warnings, and then switch on warnings as errors in CI. It may be a little bit annoying at first, but it produces dramatically better results in the long term.

## Details
- Fixed many compiler warnings. The remaining ones are overwhelmingly in the webfinger controller, and that's next to be addressed anyway.


Letterbook.Core/Models/TimelineEntry.cs
```diff
@@ -6,11 +6,23 @@ public class TimelineEntry
     // Those are often more or less the same, except in the case of boosts
     public DateTime Time { get; set; } = DateTime.UtcNow;  //index
     public ActivityObjectType Type { get; set; }
-    public string EntityId { get; set; }  //index
-    public string AudienceKey { get; set; } //index
+    public required string EntityId { get; set; }  //index
```
runewake2, line 9: I didn't know this was a thing, TIL!

runewake2 (approved):

Letterbook.Core/Models/TimelineEntry.cs
jenniferplusplus: I think it's a relatively recent thing. It might have been introduced in dotnet 6?

PR #114, opened 2023-11-05
ADR/exceptions

jenniferplusplus:
Proposal to lay the groundwork for consistent and useful error reporting


docs/decisions/04-error-reporting.md
```diff
@@ -0,0 +1,34 @@
+﻿# Subject
+
+Patterns and guidance to build consistent Exceptions with useful reporting and debugging information
+
+## Status
+
+- [ ] Decided
+- [ ] Decided Against
+- [ ] Deferred
```
jenniferplusplus, line 9:
```suggestion
- [x] Deferred (2023-12-30)
```

jenniferplusplus (commented):
Deferring this until later. I think it's important, but it may be better to revisit how to manage error reporting after we have a little operational experience.

PR #115, opened 2023-11-05
Add OTel, Prometheus and Grafana

runewake2:
Adding ASP.NET core telemetry and configuring basic prometheus + grafana endpoints. This should get us to the point where we can start recording and observing metrics, intended as a starting point for this work. By default we'll get a record of a the duration of web requests in milli's. The default behavior is discussed in more depth [here](https://github.com/open-telemetry/opentelemetry-dotnet/blob/main/src/OpenTelemetry.Instrumentation.AspNetCore/README.md).

The `provisioning/dashboards` path is kept around for future use in automatic dashboard provisioning. Implementing that now isn't super useful though since telemetry is limited.

## Details
- This introduces two new containers started w/ the project: Prometheus and Grafana. This is sensitive to the port you're running the project on and can break if that changes.
- Open Telemetry Metrics are a still an [experimental feature](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/telemetry-stability.md). There might be some turbulence as a result of this depending on how this evolves.

### DIFF
--- a/Letterbook.Api/Letterbook.Api.csproj
+++ b/Letterbook.Api/Letterbook.Api.csproj
@@ -23,6 +23,9 @@
       <PrivateAssets>all</PrivateAssets>
       <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
     </PackageReference>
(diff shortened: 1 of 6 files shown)


jenniferplusplus (approved):
love it, ship it
launch

jenniferplusplus:
Actually, just checking, you can merge PRs, right? You should be able to.

PR #121, opened 2023-11-06
Trying out APSharp

jenniferplusplus:
@warriordog I'm trying out APSharp, to see if it's viable for us at this stage. I think it is, or it's very close. This is all very impressive. But, I have some questions, so here's a PR we can use to discuss things.

A lot of the questions are inline, as comments. The big ones are:
- Is there a way to extend a base type so that it will apply to the subtypes also (Actor, in this case)? Do I need to extend each of the existing actor subtypes?
- How should I deserialize some json if I know it's an AP document, but nothing else about it? (Which will often be the case)
- Minor, but: how should I specify the context entries an extension relies on?


warriordog:
@jenniferplusplus I got your discord message about this, but I can't reply since Discord is apparently having an outage right now. 🙃 Anyway, here's the answers to your questions:

1. <ins>Is there a way to extend a base type so that it will apply to the subtypes also (Actor, in this case)? Do I need to extend each of the existing actor subtypes?</ins> Yes, but that isn't entirely implemented yet. Once everything is done, you'll be able to extend from `APActor` and then define the appropriate context and/or type name to "identity" the extension. You can then access the extended properties through the [TypeMap API](https://github.com/warriordog/ActivityPubSharp/tree/main/Source/ActivityPub.Types#usage-1). It might look like this:
```csharp
PersonActor person = Whatever();
// Work with the Person properties here

// Access the extended properties
// "personExtended" is of type YourActorExtension.
// Instances "person" and "personExtended" are tightly coupled - they share the same TypeMap internally.
    // Serializing either one will result in the exact same output.
if (person.Is<YourActorExtensions>(out var personExtended))
{
    // Work with the extended properties here

    // Just for illustration: "theSamePerson" will be the exact same instance of "person" above.
    var theSamePerson = personExtended.As<PersonActor>();
    Debug.Assert(Object.ReferenceEquals(person, theSamePerson));
}

// You can put other extensions here.
```
2. <ins>How should I deserialize some json if I know it's an AP document, but nothing else about it? (Which will often be the case)</ins> You can use [`IJsonLdSerializer`](https://github.com/warriordog/ActivityPubSharp/blob/main/Source/ActivityPub.Types/Conversion/JsonLdSerializer.cs) directly, and convert to `ASType`. You can then introspect the object in one of two ways:
```csharp
var unknownObject = jsonLdSerializer.Deserialize<ASType>(documentJson);

// Option 1: Use TypeMap API.
// This is the cleanest way and the expected usage of the library.
// Small downside: it only works for known and well-formed types.
if (unknownObject.Is<ASActivity>(out var activity))
    HandleActivity(activity); // This can further narrow the type, if desired.
if (unknownObject.Is<FireFishObject>(out var ffObject))
    NormalizeFireFishProperties(ffObject); // This is how Server Metadata feature (#129) will work
if (unknownObject.Is<LinkedDataSigned>(out var signedObject))
    ValidateLDSignature(signedObject); // Useful for middleware
// Etc. You can also use reverse these and throw if the object is *not* of a supported type.

// Option 2: Inspect the Type property.
// The library doesn't provide much support for this, but it is possible.
// I don't recommend this except as an escape hatch or if you're doing something *very* low-level.
if (unknownObject.TypeMap.ASTypes.Contains("Create"))
    HandleCreate(unknownObject);
```
3. <ins>Minor, but: how should I specify the context entries an extension relies on?</ins> I haven't actually implemented that yet. 😄 That will be completed in [issue 95](https://github.com/warriordog/ActivityPubSharp/issues/95).

Letterbook.Adapter.ActivityPub.Test/APActorExtensionsTests.cs
```diff
@@ -0,0 +1,61 @@
...
+                     "outbox": "http://letterbook.example/actor/outbox",
+                     "publicKey": {
+                       "@context": "https://www.w3.org/ns/activitystreams",
+                       "id": "http://letterbook.example/actor/keys/0",
+                       "publicKeyPem": "-----BEGIN PUBLIC KEY-----\nBase64KeyValues\\n-----END PUBLIC KEY-----\n"
+                     }
+                   }
+                   """;
+        // How would I deserialize an object whose derived type I don't know in advance? I also tried to deserialize to ASObject and APActor, both fail.
+        var poco = _serializer.Deserialize<ASType>(json); // throws
```
jenniferplusplus, line 55: @warriordog TY! I wonder if I'm defined something incorrectly, then, because this throws, and it seems like it's exactly what you're suggesting in option 1.

warriordog: What's the exception? There could be a bug in the library.

jenniferplusplus:
```
System.InvalidCastException: Can't represent the graph as type ActivityPub.Types.AS.ASType
```

warriordog: Huh, ok thanks. I'll look into that.

jenniferplusplus: If it helps, later this evening I can try to reproduce it in ActivityPubSharp and open an issue there. As opposed to using it here in letterbook from a nupkg I built myself.

warriordog:
@jenniferplusplus I'll be tracking my work on this here: https://github.com/warriordog/ActivityPubSharp/issues/131

Letterbook.Adapter.ActivityPub.Test/APActorExtensionsTests.cs
```diff
@@ -0,0 +1,61 @@
...
+        };
+
+        var json = _serializer.SerializeToElement(actor);
+        
+        Assert.True(json.TryGetProperty("publicKey", out var publicKey), "missing publicKey");
+        Assert.True(publicKey.TryGetProperty("publicKeyPem", out var publicKeyPem), "missing publicKeyPem");
+        Assert.Equal(actor.PublicKey.PublicKeyPem, publicKeyPem.GetString());
+        Assert.True(json.TryGetProperty("type", out var type));
+        
+        // It should be some kind of Actor type. How should that be set?
```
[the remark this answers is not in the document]
warriordog: FYI - this issue will simplify that redundant code: https://github.com/warriordog/ActivityPubSharp/issues/133

jenniferplusplus:
@warriordog That's awesome, thank you! I'm happy to help, starting in the next few days. I think I'm going to extend letterbook.ap for now so I can get basic federation working, because I think I'm like 1 or 2 days away from that milestone. That makes it much more approachable to start building out the baseline features and APIs for a microblogging app.

But, my preference is that we would switch to APSharp after that, rather than continue on with letterbook.ap. And we can just accept that we don't know what image blur, quotes,  reactions, and etc are until after APSharp is mature enough to support us and the extensions we want.

So, does that seem viable from your point of view?

And also, how can I help?

warriordog:
@jenniferplusplus oh, apparently I never replied to your last comment! 🤦‍♀️ I think that's a good approach. LetterBook.ActivityPub is much closer to being usable, although I think APSharp will be better in the long run (I'm biased o/c). I'll continue working on the issues we discovered here and then maybe we can circle back to this?

If you want to help, then I could use some "realistic" extension classes to test with. Your `APActorExtensions` class is a great help, and more would be even better!

Issue #116, opened 2023-11-05
Add Telemetry and Monitoring Support


runewake2:
We'd like to add telemetry and have dashboards/alerts available in app so that logs, metrics etc can be tracked easily.

- [x] Add open Telemetry Metrics support (#115)
- [x] Add logging support (#124)
- [x] Collect Postgres telemetry (#134)
- [x] Setup default dashboards and auto-provision them (#134)
- [ ] #188
- [x] Support tracing

This is an issue to help w/ tracking some of this work for getting some of this built up.

PR #125, opened 2023-11-12
Add ability to query postgres db's in grafana.

runewake2:
I don't know if we actually want this but it introduces access for both Postgres DB's in Grafana so queries can be run against tables in both of them. This allows inspecting of those DB's from inside of Grafana and potentially future use for telemetry and usage metrics for the server (ex. dashboards showing most active users).

Part of #116


jenniferplusplus (approved):
I suspect we don't want to ship this as something we have pre-configured, or even recommended. But I'm fine with it as a development tool. Grafana is a competent SQL client, and it's a good option for anyone who doesn't already have a different one they like.

runewake2:
That was my thinking, this eliminates needing to exec into the DB or setup psql in order to explore your database. Idk if that level of access is what we want for release, but it's handy to have available during dev.

Issue #126, opened 2023-11-15
Minimum Functionality project closeout/cleanup


jenniferplusplus:
Before we call it a wrap on "minimum functionality", we should clean up after ourselves a little bit. I know I've done a lot of quick and dirty work over the least few weeks. I don't want to leave that mess for everyone else to trip over.

- [x] Successful federated follows
- [x] #135
  - [x] Add Async methods to JsonLdSerializer
- [x] #137
- [x] AP client logging
- [x] Generally plan for receiving a lot of 201s and 202s from peers in the AP client
- [x] #136
- [x] Groom issues for single user preview
- [x] #138

PR #128, opened 2023-11-17
Progress on federated follows

jenniferplusplus:
This fixes some issues with following remote profiles.

## Details
- A lot of error handling
- Type mapping fixes
- A few db adapter issues
  - It turns out C# is too clever for my own good, and excludes the fragment (the part after the #) from URI equality comparisons
  - We need that part
  - Also, the generate ID thing. See the comment below.

## Related
- Progress on #126


Letterbook.Adapter.Db/EntityConfigs/ConfigureFollowerRelations.cs
```diff
@@ -8,10 +8,8 @@ public class ConfigureFollowerRelations : IEntityTypeConfiguration<Models.Follow
     public void Configure(EntityTypeBuilder<Models.FollowerRelation> builder)
     {
         builder.HasKey(relation => relation.Id);
+        builder.Property(relation => relation.Id)
+            .ValueGeneratedNever();
```
jenniferplusplus, line 12: This was a nightmare to figure out. This entity wasn't writing to the DB, but the only error was something like `The database operation was expected to affect 1 row(s), but actually affected 0 row(s)`

I've been researching and debugging it for 2 days. I eventually found this SO post with the [explanation and solution](https://stackoverflow.com/a/74932577/4100159). The problem was that we generate our own PKeys for FollowerRelations, because it's a GUID so we can. But, EFCore interpreted that as being an existing entity that we need to update, rather than a new one to insert. So it tried to do `UPDATE WHERE blah`, and matched 0 rows.

So, I learned, and now you can too.

PR #129, opened 2023-12-24
Fix remote follow

jenniferplusplus:
I made a bit of a mess with this. But! Things are working, and I'll follow up with a lot of cleanup.

The main thing:  🎉 💥  We can follow and be followed by Mastodon actors 💥 🎉 

## Details
- (Mostly) replaced letterbook.activitypub with ActivityPubSharp
  - Will finish the replacement in a follow up
  - Depends on https://github.com/warriordog/ActivityPubSharp/pull/163, update after that merges
- Adds a deferred work queue mechanism, and workers to send federated messages
  - This would eventually be necessary either way
  - But it's urgently necessary to interoperate with mastodon.

## Interop
So, interop with the rest of the fediverse. And specifically with Mastodon. Most fedi services will punt AP messages directly into a message queue and handle them later. Which is fine, I guess. It does make it hard to get any idea what's happening when things don't work. But, fine. However, this was breaking things, not just making them invisible. As far as I can tell, Mastodon doesn't commit their db transaction until the controller exits. Which means that if we send a reply message in our controller, mastodon won't know what we're talking about. We have to close the request/response cycle, and then send reply some time later. And thus, I had to build that out to make follows work, and this PR went from big and messy to huge and messy. Sorry.

## Related
- #126


Letterbook.Adapter.ActivityPub/Mappers/AsApMapper.cs
```diff
@@ -12,6 +14,32 @@ public static class AsApMapper
...
+        cfg.CreateMap<Models.Profile, PersonActorExtension>()
+            .ForMember(dest => dest.Id, opt => opt.MapFrom(src => src.Id))
+            .ForMember(dest => dest.Summary, opt => opt.MapFrom(src => src.Description))
+            .ForMember(dest => dest.Name, opt => opt.MapFrom(src => src.DisplayName))
+            .ForMember(dest => dest.PreferredUsername, opt => opt.MapFrom(src => src.Handle))
+            .ForMember(dest => dest.Attachment, opt => opt.Ignore())
+            .ForMember(dest => dest.Liked, opt => opt.Ignore())
+            .ForMember(dest => dest.PublicKey,
+                opt => opt.MapFrom<SigningKeyConverter, IList<SigningKey>>(src => src.Keys))
+            .ForMember(dest => dest.Endpoints, opt => opt.Ignore())
```
warriordog, line 33: Are endpoints intentionally ignored?

Letterbook.Adapter.ActivityPub/Mappers/SigningKeyConverter.cs
```diff
@@ -48,4 +52,50 @@ public AsAp.PublicKey Convert(Models.SigningKey source, AsAp.PublicKey? destinat
         destMember.Owner = context.Mapper.Map<AsAp.Link>(source.Id);
         return destMember;
     }
+
+    // TODO(now): for APSharp
```
warriordog, line 56: this TODO is marked as "now". should it be implemented?

Letterbook.Adapter.ActivityPub/Mappers/SigningKeyConverter.cs
```diff
@@ -48,4 +52,50 @@ public AsAp.PublicKey Convert(Models.SigningKey source, AsAp.PublicKey? destinat
...
+        return destMember;
+    }
+
+    public PublicKey? Convert(IList<Models.SigningKey> source, PublicKey? destination, ResolutionContext context)
+    {
+        var signingKey = source.OrderBy(key => key.KeyOrder)
+            .FirstOrDefault(key => key.Family == Models.SigningKey.KeyFamily.Rsa);
+        if (signingKey == null)
+        {
+            return default!;
```
warriordog, line 70: this null-forgiving operator may be redundant

Letterbook.Adapter.ActivityPub/Mappers/SigningKeyConverter.cs
```diff
@@ -48,4 +52,50 @@ public AsAp.PublicKey Convert(Models.SigningKey source, AsAp.PublicKey? destinat
...
+        var pem = source.Family switch
+        {
+            Models.SigningKey.KeyFamily.Rsa => source.GetRsa().ExportSubjectPublicKeyInfoPem(),
+            Models.SigningKey.KeyFamily.Dsa => source.GetDsa().ExportSubjectPublicKeyInfoPem(),
+            Models.SigningKey.KeyFamily.EcDsa => source.GetEcDsa().ExportSubjectPublicKeyInfoPem(),
+            _ => null
+        };
+        if (pem == null)
+        {
+            return default!;
```
warriordog, line 96: same as previous comment

Letterbook.Adapter.ActivityPub/Types/PersonActorExtension.cs
warriordog, line 1: These extension classes need a small tweak to support the latest build of APSharp. I can submit a separate PR for that, if you'd like.

Letterbook.Adapter.RxMessageBus/RxMessageChannels.cs
```diff
@@ -0,0 +1,31 @@
...
+    
+    public RxMessageChannels(ILogger<RxMessageChannels> logger)
+    {
+        _logger = logger;
+        _channels = new Dictionary<string, Subject<CloudEvent>>();
+    }
+    
+    public Subject<CloudEvent> GetSubject(string type)
+    {
+        if (!_channels.ContainsKey(type))
```
warriordog, line 20: suggestion: use TryGetValue

Letterbook.Api/Controllers/ActivityPub/ActorController.cs
```diff
@@ -17,26 +25,27 @@ namespace Letterbook.Api.Controllers.ActivityPub;
 /// </summary>
 [ApiController]
 [Route("[controller]")]
-[AcceptHeader("application/ld+json",
+[Consumes("application/ld+json",
     "application/ld+json; profile=\"https://www.w3.org/ns/activitystreams\"", 
     "application/activity+json")]
-[JsonLdSerializer]
+// [JsonLdSerializer]
```
warriordog, line 31: commented code

Letterbook.Api/JsonLdSerializerAttribute.cs
```diff
@@ -12,9 +13,9 @@ public override void OnActionExecuted(ActionExecutedContext ctx)
         if (ctx.Result is OkObjectResult objectResult)
         {
             objectResult.Formatters.Clear();
-            objectResult.Formatters.Add(new SystemTextJsonOutputFormatter(JsonOptions.ActivityPub));
-            objectResult.ContentTypes.Clear();
-            objectResult.ContentTypes.Add("application/ld+json");
+            objectResult.Formatters.Add(new JsonLdOutputFormatter());
+            // objectResult.ContentTypes.Clear();
```
warriordog, line 17: commented code

Letterbook.Core/AccountEventService.cs
```diff
@@ -16,7 +16,7 @@ public AccountEventService(IOptions<CoreOptions> options, IMessageBusAdapter mes
     {
         _options = options.Value;
         _messageBusAdapter = messageBusAdapter;
-        _channel = _messageBusAdapter.OpenChannel<Account>();
+        _channel = _messageBusAdapter.OpenChannel<Account>(nameof(ActivityEventService));
```
warriordog, line 19: This is in AccountEventService, but uses ActivityEventService. Is that intentional?

Letterbook.Core/Adapters/IActivityPubAuthenticatedClient.cs
warriordog, line 1: suggestion: document why some of these return untyped `object`

Letterbook.Adapter.ActivityPub.Test/MapperTests.cs
```diff
@@ -14,19 +15,22 @@ public class MapperTests
...
-    
+
 
     public MapperTests(ITestOutputHelper output)
     {
         _output = output;
         _profileMapper = new Mapper(ProfileMappers.DefaultProfile);
         _asApMapper = new Mapper(AsApMapper.Config);
+        _APSharpMapper = new Mapper(AsApMapper.ActorConfig);
         // _asApActor = new Mapper(AsApMapper.DefaultActor);
```
warriordog, line 32: commented code

warriordog (approved):
@jenniferplusplus awesome work with this! What a crazy and difficult journey just to follow an account!

Letterbook.Adapter.ActivityPub/Document.cs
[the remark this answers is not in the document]
jenniferplusplus: It could, but that would require core classes to take a direct dependency on the .ActivityPub project, and I would rather they get it from dependency injection.

Letterbook.Adapter.ActivityPub/Extensions.cs
```diff
@@ -0,0 +1,56 @@
...
+        {
+            id = link.HRef;
+            return true;
+        }
+
+        id = null;
+        return false;
+    }
+
+    public static bool TryGetId(this ASObject aso, [NotNullWhen(true)]out Uri? id)
```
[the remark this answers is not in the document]
jenniferplusplus: Yes, according to the spec. But given the state of all the existing implementations, it didn't seem unreasonable to be permissive about that

https://www.w3.org/TR/activitystreams-vocabulary/#dfn-id

Letterbook.Adapter.ActivityPub/Mappers/AsApMapper.cs
jenniferplusplus: This is mostly copied from the letterbook.ap mapper, and I hadn't implemented endpoints. So, probably not.

Letterbook.Adapter.ActivityPub/Client.cs
```diff
@@ -83,52 +130,23 @@ private HttpRequestMessage SignedRequest(HttpMethod method, Uri uri)
...
-        {
-            var responseActivity = await JsonSerializer.DeserializeAsync<AsAp.Activity>(stream, JsonOptions.ActivityPub);
-            if (responseActivity == null) throw new NotImplementedException();
-
-            if (responseActivity.Type.Equals("Accept", StringComparison.InvariantCultureIgnoreCase))
-                return FollowState.Accepted;
-        
-            if (responseActivity.Type.Equals("Reject", StringComparison.InvariantCultureIgnoreCase))
-                return FollowState.Rejected;
+        var doc = _document.Follow(_profile!, target);
```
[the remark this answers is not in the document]
jenniferplusplus: Yeah, this method only exists on the AuthenticatedClient interface. I should maybe split those into different classes or something so I don't have to do null forgiveness like this all the time.

jenniferplusplus:
@warriordog ty for the review! I still have a *ton* of cleanup to do, obviously. I'll do that and then revisit the rest of your comments.

Letterbook.Adapter.ActivityPub/Extensions.cs
```diff
@@ -0,0 +1,56 @@
+﻿using System.Diagnostics.CodeAnalysis;
+using ActivityPub.Types.AS;
+using ActivityPub.Types.Util;
+
+namespace Letterbook.Adapter.ActivityPub;
+
+public static class Extensions
+{
+    public static bool TryGetId(this Linkable<ASObject> linkable, [NotNullWhen(true)]out Uri? id)
```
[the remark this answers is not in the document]
jenniferplusplus: It's not obvious how I would do that. Linkable<> doesn't inherit from ASType, or anything for that matter.

Letterbook.Adapter.ActivityPub/Types/PersonActorExtension.cs
jenniferplusplus: I think I got it, but I'm happy to have you check again now that it's updated

Letterbook.Adapter.ActivityPub/Extensions.cs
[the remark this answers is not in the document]
jenniferplusplus: @warriordog actually, I think I spoke too soon. It's probably spec compliant (but incredibly obnoxious) for `id` to be a compact IRI, too.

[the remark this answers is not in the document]
warriordog: @jenniferplusplus oh now I remember, compact IRIs are actually why I kept it as a string

Letterbook.Adapter.ActivityPub/Extensions.cs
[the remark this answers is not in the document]
warriordog: Something like this:
```suggestion
    public static bool TryGetId<T>(this Linkable<T> linkable, [NotNullWhen(true)] out Uri? id)
    where T : ASObject
```

Letterbook.Adapter.ActivityPub/Document.cs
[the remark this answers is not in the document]
warriordog: makes sense

warriordog:
@jenniferplusplus no problem at all!

Issue #139, opened 2023-12-27
Media features


jenniferplusplus:
File and media services would allow us to support rich media. That is to say: images, video, and audio. 

## Need
- [ ] Update Image model, if necessary
- [ ] Add core file and/or media services to create and manage media
  - [ ] Be sure to emit events related to things happening with media. We'll likely need a significant amount of background processing for it
- [ ] Add db adapter queries for media
- [ ] Add file adapter
  - [ ] Might be good to support filesystem storage
  - [ ] But the expected path is S3-compatible object storage
- [ ] Add API controllers and actions related to files and media
- [ ] Handle ActivityPub media objects like Image and Video

> [!TIP]
> Update above list as needed, and convert them to issues when they're understood well enough to start development

## Goal
The basic goal is to be able to upload, store, manage, and serve common image formats.

Keep audio, video, and *maybeeee* documents in mind when working on this, but there's no need to implement them, yet. Audio and video are very technically complex. They can also raise particular trust & safety concerns, and can be hard or taxing to moderate. Let's not try to support them right away.

We'll eventually need to do automated scanning, fingerprinting, re-encoding, and maybe other things for audio. But, that's all tremedously complex, and we can start small. Just keep those things in mind, and try to produce events we can utilize to hook into the process later on.

PR #144, opened 2023-12-30
ADR-07: post data model

jenniferplusplus:
Proposes a change to the content data model, which should make both inbound activity processing and general maintenance easier. It's an ADR, so the ADR doc itself is really the best description.
https://github.com/Letterbook/Letterbook/compare/adr/post-data-model?expand=1&short_path=50f4b27#diff-50f4b271e491af84f4eaadb6740d8ca218ff6a810c2441bab15fb76eab205958

## Details
- Introduce the concept of a post
- Concepts like Notes and Images then become content within a post

## Related
- #139 
- #126

### DIFF
--- a/Letterbook.Core/Models/Post.cs
+++ b/Letterbook.Core/Models/Post.cs
@@ -1,0 +1,25 @@
+﻿namespace Letterbook.Core.Models;
+
+public class Post
+{
+    public Uri Id { get; set; }
+    public Guid? LocalId { get; set; }
+    public string Authority => Id.Authority;
+    public ICollection<Profile> Creators { get; set; } = new HashSet<Profile>();
+    public DateTime CreatedDate { get; set; }
+    public ICollection<IContent> Contents { get; set; } = new HashSet<IContent>();
+    public ICollection<Audience> Visibility { get; set; } = new HashSet<Audience>();
+    public ICollection<Mention> AddressedTo { get; set; } = new HashSet<Mention>();
+    public string? Client { get; set; }
+    public Post? InReplyTo { get; set; }
+    public Uri? Replies { get; set; }
+    public IList<Post> RepliesCollection { get; set; } = new List<Post>();
+    public Uri? Likes { get; set; }
+    public IList<Profile> LikesCollection { get; set; } = new List<Profile>();
+    public Uri? Shares { get; set; }
+    public IList<Profile> SharesCollection { get; set; } = new List<Profile>();
+}
+
+public interface IContent
+{
+}
--- a/Letterbook.Core/Models/Post.cs
+++ b/Letterbook.Core/Models/Post.cs
@@ -8,7 +8,7 @@ public class Post
     public ICollection<Profile> Creators { get; set; } = new HashSet<Profile>();
     public DateTime CreatedDate { get; set; }
     public ICollection<IContent> Contents { get; set; } = new HashSet<IContent>();
-    public ICollection<Audience> Visibility { get; set; } = new HashSet<Audience>();
+    public ICollection<Audience> Audience { get; set; } = new HashSet<Audience>();
     public ICollection<Mention> AddressedTo { get; set; } = new HashSet<Mention>();
     public string? Client { get; set; }
     public Post? InReplyTo { get; set; }
--- a/docs/decisions/07-posts-data-model.md
+++ b/docs/decisions/07-posts-data-model.md
@@ -1,6 +1,6 @@
 ﻿# Data Model for Generic Posts
 
-We can expect almost any of the ActivityStreams Object types to federate in a way that resembles a post. And we will want to do the same eventually; we don't need to limit ourselves to just Notes. This offers a way to represent posts in a timeline, thread, etc as a generic container for their content.
+We can expect almost any of the ActivityStreams Object types to federate in a way that resembles a post. And we will want to do the same eventually; we don't need to limit ourselves to just Notes. This offers a way to represent posts in a timeline, thread, etc independently of their specific content.
 
 ## Status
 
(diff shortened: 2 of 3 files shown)


Letterbook.Core/Models/Post.cs
```diff
@@ -0,0 +1,25 @@
...
+    public Post? InReplyTo { get; set; }
+    public Uri? Replies { get; set; }
+    public IList<Post> RepliesCollection { get; set; } = new List<Post>();
+    public Uri? Likes { get; set; }
+    public IList<Profile> LikesCollection { get; set; } = new List<Profile>();
+    public Uri? Shares { get; set; }
+    public IList<Profile> SharesCollection { get; set; } = new List<Profile>();
+}
+
+public interface IContent
```
runewake2, line 23: The ADR defines some contents for this interface, should that be present here as well?

docs/decisions/07-posts-data-model.md
```diff
@@ -0,0 +1,162 @@
...
+        Type()*
+    }
+
+    Content <|-- Note : inherits
+    Content <|-- Image : inherits
+    Content <|-- Etc : inherits
+    IContent <|-- Content : implements
+    Post --* IContent : contains
+    Post --> FeedDto : maps-to
+    FeedDto --* Content : contains
```
runewake2, line 127: I'm curious how would you see Quotes or Citation posts fitting into this model? Those types of content seem to lean into the graphical representation this wants to move away from.

docs/decisions/07-posts-data-model.md
```diff
@@ -0,0 +1,162 @@
...
+```
+
+## Impact
+
+This will obviously introduce some new data types. It's also likely to require some complex migrations. It should ideally happen soon, before any serious work starts on posting-related features.
+
+You may also notice that this data model is more restrictive than what ActivityPub would allow. Notably, Posts can only be InReplyTo at most one other Post (and no other kind of object). A Note can only be in one Post and vice-versa. And because `IContent.Thread` would map to `AS Context`, a post can only have one context and exist in one thread. These restrictions should be fully compatible with the actual behavior of other fedi services in the wild. It helps to narrow the possibility space that we would actually have to deal with, and should support faster queries for the most common cases (when a post is a note and nothing else). This means that there could be potential future fedi peers that we would not be readily compatible with.
```
runewake2, line 134: > A Note can only be in one Post and vice-versa.

Does this align with the discussion below about a content item theoretically existing in multiple posts?

docs/decisions/07-posts-data-model.md
```diff
@@ -0,0 +1,162 @@
...
+### schema
+Making each content record include-able in only one post saves on potentially a _ton_ of many-many joins. This makes it more involved to re-use uploaded media. For example, to reuse the same reaction gif repeatedly. We would need to introduce other models to track those uploads as files, not just post content. But, that seems worth it to me.
+
+We can defer that model design until we start work on the [file/media features](https://github.com/Letterbook/Letterbook/issues/139).
+
+### hypothetical compatibility
+Not attempting to be universally future compatible seems reasonable. It's not like it would be possible anyway. We can work to improve compatibility in the future, if such an incompatible peer service should ever be created.
+
+### need the interface?
+Do we need `IContent`, or is the base/abstract `Content` class enough?
```
runewake2, line 159: There's potentially some testing benefit via mocking? But I'm not sure where or why we'd need that? Why do we need an abstract `Content` though - seems like default implementations and extension methods would cover most of the needs of the general class? Maybe I am overlooking something?

docs/decisions/07-posts-data-model.md
jenniferplusplus: That's a fair point. I wish it was easier to find info about how other software federates quote posts. What I don't want to do is the mastodon thing of just having the one kind of post, which is basically a note with media attachments. So, I was kind of putting that off a little. The best thing might be if `Post` also implements `IContent`, so that the content list can also contain other posts.

Other options I can think of:
1. Note and Post are the same concept, and post gets a `public Post Quote` field.
2. Add a `public Post Quote` to the Note type, and future types that could meaningfully embed a quote.

docs/decisions/07-posts-data-model.md
jenniferplusplus: I think so. I meant for it to. The ADR is mostly about the internal data model. But this specific paragraph is about activitypub messages, and how this theoretically narrows the message compositions we can understand. But I don't think any existing implementation is actually doing any of the things this would exclude, so it's really just theoretical.

docs/decisions/07-posts-data-model.md
jenniferplusplus: Yeah, there may not be much benefit to making it abstract. My thinking there was mostly that I don't want that base class to accidentally end up in the relational db schema.

docs/decisions/07-posts-data-model.md
```diff
@@ -0,0 +1,162 @@
...
+
+## Discussion
+
+### minor activities
+A Post can have more than one content, and a content item can theoretically exist in more than one post. If a post is `Liked` by someone, should that Like follow the content to other posts? I think probably not, right? For example, suppose you upload a photo, then post that photo with a note. The later you add that photo to an album. Should the view of the photo in the album show all the Likes that the Post received? What if you reuse the photo in another post in the future?
```
jenniferplusplus, line 145:
```suggestion
### minor activities
A Post can have more than one content, and a content item can theoretically exist in more than one post. If a post is `Liked` by someone, should that Like follow the content to other posts? I think probably not, right? For example, suppose you upload a photo, then post that photo with a note. The later you add that photo to an album. Should the view of the photo in the album show all the Likes that the Post received? What if you reuse the photo in another post in the future?
```

Ignore this, it's only relevant to an earlier iteration that I think was too complex and might have presented performance concerns

Letterbook.Core/Models/Post.cs
jenniferplusplus: I included the code change mostly for illustration. The whole implementation is probably a few days work.

PR #152, opened 2024-01-06
In HTTP signature validation, compare key IDs as URIs

IsaacSchemm:
When validating an HTTP signature, the `Uri` representing the actor's key ID is converted to a string and then compared to the incoming key ID. This pull request replaces this check with a conversion of the incoming key ID into a `Uri` and then a comparison against the actor's key ID `Uri`.

## Details
Single-actor servers like [microblog.pub](https://docs.microblog.pub/) may use actor and key IDs whose URIs are at the root of a domain, e.g. `https://microblog.lakora.us#main-key`. When these IDs are parsed into a `Uri`, the `ToString` method adds a trailing slash, e.g. `https://microblog.lakora.us/#main-key`, which causes the comparison to fail.

By parsing the incoming key ID as a `Uri` instead, we can compare the two `Uri` objects, which will disregard any such trailing slash.

An alternate solution would be to compare to `OriginalString` instead of `ToString()`, which would enforce the presence or absence of the trailing slash in the incoming key ID, but preserve whether or not it was present in the actor object.

## Related
I've used Letterbook's HTTP signature validation code (with this change, and some removals) in my hobby project [Crowmask](https://github.com/IsaacSchemm/Crowmask) and it's able to talk to my microblog.pub server at https://microblog.lakora.us.


jenniferplusplus:
@IsaacSchemm that's helpful, thank you! Would you mind adding a test that captures this issue? Once the basic app features are more built out, I intend to revisit http signatures. I eventually want to split it out to a dedicated package, and I want it to support more modern drafts of the spec (or the finalized version if it finalizes). It would be important to have tests when we get to that point.

IsaacSchemm:
I've added a test for the cases with and without the trailing slash. I included a hardcoded signature because the signer in the application currently always signs it with the trailing slash (which is fine, it just means it can't be used to test the case without one).

jenniferplusplus:
@IsaacSchemm Thanks, I appreciate it

PR #160, opened 2024-01-26
Account creation

### LINKED ISSUE #158: Update account management core services
- Account creation
- Account management
  - Manage email(s?)
  - Other stuff?
- Login/Logout

jenniferplusplus:
Handle account creation, at least in the simple case.

> [!WARNING]
> Migrations reset, you'll need to drop your local db, or manually correct the migrations log

## Details
- Needs better error handling, duplicate checking, actual event messaging and downstream stuff
- blah, it's usable
- go forth and build
- I discovered a problem with migrations that were a couple migrations old. It's not worth spending the time to fix it right now.

## Related
- closes #158


runewake2:
Quick look, this looks good. I can spend a bit more time tonight looking at it.

Are you expecting to change how we return account registration errors to the user in the future? ex. Not throwing errors from the UserManager all the way back to the public API's?

jenniferplusplus:
@runewake2 generally, yes. I can add some of that tonight. The returned `IdentityResult` captures a lot of failure scenarios and attaches them as an `Errors` list. But it can throw, and I should handle that better.

PR #163, opened 2024-01-28
Internals/PKeys -> UUIDv7

### LINKED ISSUE #162: Use UUIDv7 as pkey

jenniferplusplus:
Make (nearly) all of the primary keys into UUIDv7 values. UUIDv7 is a time-ordered UUID. They're meaningfully sortable, and still nearly as collision-resistant as v4. This is great for a lot of reasons:

## Benefits
- vs using URIs, like we had been in many places: way way better indexing performance. Like, 10x.
- vs using Guid, like we had been as secondary indexes:
  - considerably better index performance, because related rows will tend to have similar key values, so they're much more likely to be in the same b-tree bucket, meaning fewer searches through the index.
  - They're sortable. Records have a natural order. That makes it *way* easier to produce stable ordering in query results, which is super critical for paged queries.
- vs using ShortId: The 25 char representation that this library offers has better memory performance, and the 25 char string is still sortable.

## Related
- closes #162 
- This is the kind of change that makes me feel good about the amount of test coverage we have. Sometimes I worry it's too low, but all the really important bits are well covered.


Letterbook.Core.Tests/DeliveryWorkerTests.cs
```diff
@@ -24,33 +27,35 @@ public class DeliveryWorkerTests : WithMocks, IClassFixture<JsonLdSerializerFixt
...
+    public DeliveryWorkerTests(ITestOutputHelper output, JsonLdSerializerFixture serializer,
+        TestOutputLoggerFixture outputLogger)
     {
         _output = output;
+        _outputLogger = outputLogger;
         _output.WriteLine($"Bogus seed: {Init.WithSeed()}");
-        _mockWorker = new Mock<DeliveryWorker>(Mock.Of<ILogger<DeliveryWorker>>(), AccountProfileMock.Object,
-            ActivityPubClientMock.Object);
+
+        _mockWorker = new Mock<DeliveryWorker>(outputLogger.CreateLogger<DeliveryWorker>(output),
```
jenniferplusplus, line 37: Actually, I think this is really useful to highlight.

OutputLogger creates an ILogger<T> that can be injected into systems under test, and which is backed by the XUnit ITestOutpuHelper. This makes it straightforward in most cases to get the service's logs in the test output. Possibly very helpful in debugging test failures. Especially in message workers like this one.

PR #170, opened 2024-02-17
Adr-08 architecture update

jenniferplusplus:
Docs for the current state of the software architecture. That makes this a great time to ask questions or make suggestions.

[Jump right to the interesting part](https://github.com/Letterbook/Letterbook/pull/170/files?short_path=2a734db#diff-2a734dbaf2b6852b3d3a258b2d448fa4b22e0d2fb8c0b6dcd53e27376abda617)

cc @mattly


docs/decisions/08-architecture-update-01.md
```diff
@@ -0,0 +1,136 @@
...
+    Api -- > d
+    Api -- > f
+```
+-->
+![architecture-update-01.svg](assets%2Farchitecture-update-01.svg)
+*not implemented yet
+
+So, generally, core logic services sit in the center of the architecture, and everything else depends on them, or implements an interface that core services depend on. This isolates the core logic from any concerns related to interacting with outside systems or infrastructure. And the adapters are agnostic to the specifics of the core logic. They will generally handle core data models, but mostly to serialize them. Just like with other dependencies, core services should only handle core data models. They should generally not handle serialization or DTOs.
+
+## Core Services
```
warriordog, line 104: suggestion: define / document the process for resolving overlap (when a particular concern fits equally into two or more existing services)

docs/decisions/08-architecture-update-01.md
```diff
@@ -0,0 +1,136 @@
...
+* `AuthorizationService` encapsulates authorization logic for the rest of the application.
+
+In addition, there are a variety of event and message services. These are generally pretty thin, and just exist to consistently schedule messages to an adapter.
+
+## Infrastructure
+
+* Postgresql is the core database for Letterbook. This is the system of record, and it stores all of Letterbook's application data.
+* Timescale is a specialty timeseries database. Letterbook uses it to provide a materialized view of post feeds (and soon, notifications). Timescale is actually a plugin for Postgres, so in many cases it's possible to run with only the one database server.
+
+These are not yet implemented, but Letterbook will also depend on an S3-compatible object store, and an email service. Farther down the line, it's also likely we'll optionally incorporate a text search database like ElasticSearch.
```
warriordog, line 118: > Letterbook will also depend on an S3-compatible object store

Is this going to be a hard requirement, or just a supported feature?

warriordog (approved):
@jenniferplusplus great work with the documents! I left a couple of comments, but they're minor things.

docs/decisions/08-architecture-update-01.md
jenniferplusplus: At a technical level, it's not difficult to use the local filesystem instead of object storage. In fact, it's probably easier. But, I think it's setting yourself up to have whole classes of operational problems that don't need to exist. So, I'm really reluctant to encourage that kind of configuration.

warriordog: ok. I have a lot of thoughts on that topic, but that should be a separate discussion.

jenniferplusplus: To summarize an out of band conversation, for posterity:

Object storage isn't ideal in all situations, and some operators may prefer to just use the local filesystem.
1. Cloud object storage can feel unnecessarily complicated for small servers.
2. It can also be hard to control billing on cloud services.

Personally, I think that's unwise because it creates large classes of problems that otherwise don't exist
- it presents a DoS vector from bad actors just filling your disk
- it limits the reliability and performance to whatever your server's local disks can support
- the instance operators now have to manage filesystem backups
- it makes the server itself stateful
    - it's no longer simple to add capacity by putting a 2nd instance behind a load balancer
    - it complicates server maintenance, because servers can't be treated as ephemeral

It is of course possible to work around those issues. But at that point, I think you might as well be using object storage. Still, I expect we will implement storage on the local filesystem at some point, because it's borderline trivial to do so.

PR #171, opened 2024-02-25
Remove this dangling folder include

AnthonySteele:
What is the purpose of this PR? What does it do, and how?

Remove this dangling folder include for `Fixtures` from `Letterbook.Adapter.ActivityPub.Test.csproj` The folder does not exist.
If it did, it would be included by default, There is in most cases no need to manually include a folder
This is likely cruft, unless there is a specific reason for it (such as generated code?)

## Details
- Are there any technical details you need to explain?
- New patterns, or deviations from existing patterns?

## Related
- link issues, this helps when preparing release notes
- use closing keywords, if appropriate

### DIFF
--- a/Letterbook.Adapter.ActivityPub.Test/Letterbook.Adapter.ActivityPub.Test.csproj
+++ b/Letterbook.Adapter.ActivityPub.Test/Letterbook.Adapter.ActivityPub.Test.csproj
@@ -42,8 +42,4 @@
       </None>
     </ItemGroup>
 
-    <ItemGroup>
-      <Folder Include="Fixtures\" />
-    </ItemGroup>
-
 </Project>


jenniferplusplus (approved):
Thanks!

FYI, there are some problems with running from main right now. It's related to EFCore's handling of our UUIDv7 PKeys. I'm wrapping up the fixes for that right now, but it will be a really big merge. Just something to be aware of, if you're doing anything more substantial.

(Also, if you're doing anything more substantial, I'd be happy to talk through it, so we're not stepping on each other's toes.)

AnthonySteele:
> It's related to EFCore's handling of our UUIDv7 PKeys.

Right, I had issues building the integration tests solution, which must be related to that.

I'm planning to start with small, obvious infrastructural things such as this before attempting anything more complex.
There is 1 more like this: https://github.com/Letterbook/Letterbook/pull/173
I'm thinking of running `snitch` to check package versions
And looking at the Github actions node version warnings

AnthonySteele:
Past that, I might go down the route of `.editorconfig` and removing warnings etc.

Or looking at how to drive a feature from an outside-in test.

or both.

AnthonySteele:
Both solutions build for me now.

I want to re-iterate that the TestHost is not just for "Integration Tests", it is useful for "unit testing" and driving expected behaviour from the outside in, in a test-first way.

jenniferplusplus:
I think we have different testing philosophies. I want narrow unit tests that are strictly decoupled because I want the application software to be designed the same way. The tests then become a development tool. We can use them to explore, debug, and describe the behavior of individual components. It also makes tests fast, idempotent, and easy to manage, because they have no persistent state.

I'm not likely to turn away additional testing, but that's really not something I'm looking for at this stage.

jenniferplusplus:
@AnthonySteele We have a small discord for contributors. It can help to have detailed technical discussions more in real time, rather than in comment threads. I can send you and invite on mastodon (or anywhere you prefer, but your mastodon handle is in your gh profile). Let me know if you'd like that.

AnthonySteele:
I see tests that are strictly decoupled from the implementation details of the code under test as a good development tool, because it's mostly at the level of interactions with the running program's public API, and thus is at the level of the expected features, not the class structures and method signatures that mocks force you into.
 
 I want to re-iterate that what I am saying is that the TestHost is not just for "tests that have persistent state" but for idempotent  unit tests.

PR #174, opened 2024-02-26
Cleanup unused imports

jenniferplusplus:
Over time, some packages snuck into our imports that we don't actually use. This removes them.

## Details
- Newtonsoft
- JsonDiffPatch
- NJsonSchema

(huh, it's all json)


AnthonySteele:
There is a dotnet command line tool called [snitch](https://github.com/spectresystems/snitch) that is good at finding u used package refs

PR #179, opened 2024-03-02
Add a basic .editorconfig file

AnthonySteele:
What is the purpose of this PR? What does it do, and how?

Add a basic .editorconfig file

## Details

* covers only `.cs` files
* Covers only basic formatting: 
  * utf-8
  * 4 spaces not tabs
  * No trailing newline at end of file (simply cause that's what the existing files look like)
  * No trailing whitespace at end of line
  * LF / CRLF not configured, as in my experience it's better to let Git take care of that and not have the two tools fight over it, that does not end well. 

There will be initial changes because of this when files are worked on,  specifically the " trailing whitespace at end of line" and utf-8 options. But whatever option you pick, that will happen as files fall into line

It is possible to extend this, in 2 directions
- there are loads of [dotnet code-style rule options](https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/code-style-rule-options) that are supported by some dotnet tools.
- cover other file types that are used in the repo.

But these are not needed initially.

## Related

https://editorconfig.org/

https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/code-style-rule-options

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,9 @@
+root = true
+
+[*.cs]
+charset = utf-8
+indent_style = space
+indent_size = 4
+max_line_length = 240
+insert_final_newline = false
+trim_trailing_whitespace = true


jenniferplusplus:
Thanks for doing this. I have 2 thoughts. First, 240 line length seems really long. That's bound to introduce horizontal scrolling for almost any setup.

Second, it seems like the consensus is that tabs are more accessible than spaces. Personally, I just want to let jetbrains do whatever it decides to do. But if we're making explicit choices, then we should choose the one that's easier for disabled people to adapt to their needs.

AnthonySteele:
I would guess that lines currently have no max length at all that format tools use, so are essentially unlimited if you chose to write code that bad. My IDE on wide screen can accommodate 140 chars or more - the days of 80 columns are over. I'm happy to set it to a lower value such as 120 or 140? This is more a backstop than anything else.

With spaces, I was going by what I saw in the code - it's all 4 spaces. It can be set to tabs, sure and I do recall reading that this is better for accessibility. But for warning, I think that should be followed up by a "change all spaces everywhere" PR and some find that kind of commit annoying in the history.

jenniferplusplus:
140 is probably good.

And yes, we would need some sweeping changes to update formatting. But I think now is the time for that. It's only going to become more disruptive as the project continues. I can do that, if you'd rather not.

jenniferplusplus:
If you do make that tab style change now, would you mind including an ADR doc to that effect? That way we have something to refer back to later.

jenniferplusplus (approved):

jenniferplusplus:
LGTM, happy to merge, unless you have more coming

AnthonySteele:
In the spirit of incremental progress, I am happy to merge this now. 
I did play around with additional settings `indent_size` and `tab_width` but they don't seem to be necessary (and therefor we're better of without them as extras). 

I think that 1 benefit of tabs and not setting those size/with options is that if e.g. you want the tabs to be shown as an indent of 2 spaces per tab, and I need them to be an indent of 6, then we can both have this view as our display settings _without changing a single byte of any .cs file_.

But as discussed this change likely means that soon you should close off existing PRs, then run a mass reformat. Probably better if you do that, at a time of your choosing.

Issue #35, opened 2023-07-05
Authentication


jenniferplusplus:
Some kind of login flow for users. I will take any solution right now, but eventually I'd like to use OIDC.

jenniferplusplus:
Wow doing this in dotnet sucks right now. [Supposedly it's going to get a lot better in dotnet 8](https://devblogs.microsoft.com/dotnet/improvements-auth-identity-aspnetcore-8/). I hope so. For now, I think I'm going to punt on this and come back to it after I've made some progress on some other things.

If anyone wants to give it a shot in the meantime, be my guest

jenniferplusplus:
I'm putting together a solution involving AspNet Identity and OpenIddict. This doesn't seem to actually be all that much easier in dotnet 8, so I'm dropping that for now.

PR #183, opened 2024-03-07
ADR-11 Authorization

jenniferplusplus:
Proposes an attribute based authorization scheme.

## Details
- Formal claims would be very significant attributes within this scheme
- Presents an example set of claims to base work from
- Presents some examples of how those claims could be encoded in authz tokens

## Related
- #212 
- #36 
- #35


docs/decisions/08-abac-claims.md
```diff
@@ -0,0 +1,98 @@
...
+  - `ReadDraft`
+  - `React`
+  - `Share`
+  - `SortReply`
+  - `ApproveReply`
+  - `Collect`
+  - Coauthor stuff?
+  - Mentions?
+  - Private mentions?
+  - Media stuff?
```
jenniferplusplus, line 49: I'm not sure how these should be broken down, and maybe it's best to defer it to later? Coauthoring is a feature I really want to support, but it's going to be a while. Mention-related attributes seem like they would be helpful to moderators. You could, for instance, revoke someone's ability to send mastodon-style private messages, which I expect could curtail some spam and abuse.

But, maybe it should wait until we're actually building out moderation features?

docs/decisions/08-abac-claims.md
```diff
@@ -0,0 +1,98 @@
...
+```json
+{
+  "kid": "blah",
+  "alg": "blah"
+}.{
+  "iss": "https://letterbook.example",
+  "exp": 1704088800,
+  "profiles": {
+    "coffee_nebula": ["Owner"],
+    "seven9": ["ApproveReply", "SortReply", "ApproveFollower", "Block", "Guest"]
```
jenniferplusplus, line 79: Thoughts on nested claims like in the doc, vs namespaced ones like 
```json
"profile:coffee_nebula": [],
"profile:seven9": [],
"account:katherine_janeway": []
```

jenniferplusplus: also, to be clear, these would all be UUIDs (or id25 representations of UUIDs). I used words here for readability

jenniferplusplus: Yeah. It is unbounded, but as a practical matter, I think it would be rare to have claims on *any* profile but your own, and very rare to have more than a handful of them. For accounts, I expect it would be exceedingly rare to have claims on any but your own. In fact, the only reason I wanted to support claims on multiple accounts in a token is to facilitate admins and moderators being able to look up contact info.

Perhaps supported by a wildcard entry? Like `"profile:*": ["Read"], "account:*": ["ViewInfo"]`? And that could possibly be done via an explicit and sort lived privilege escalation.

docs/decisions/08-abac-claims.md
```diff
@@ -0,0 +1,98 @@
...
+```
+In this example, this token was issued to `katherine_janeway`, the owner of profile `coffee_nebula`, and has also been granted some claims on profile `seven9`. Those additional claims would allow Janeway to assist with performing moderation tasks on the replies to posts made by `seven9`. This kind of thing is really helpful for the targets of internet pile-ons, and this authorization scheme allows us to facilitate that. In other apps, they would have had to just share passwords, which is bad for a lot of reasons.
+
+## Non-claim attributes
+
+Not all attributes can be expressed as claims on an auth token. For one thing, it's not practical. Tokens are usually passed as HTTP headers, and there are limitations on header size. For another, we may just not want to represent them as claims. For example. Members of a group should be able to read messages presented to the group. But it's possible that profiles could join and leave many groups, somewhat frequently. Claims attached to authentication tokens would need to trace back and update account-level models in order to express that, and that seems unwise. So, some attributes will need to be deduced from run-time data.
```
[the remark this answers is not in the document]
jenniferplusplus: Yeah, I certainly think that seconds is too short a lifetime in order to gain any real benefit from embedding claims into tokens. I was imagining something in the neighborhood of 10 minutes.

Re: coauthoring specifically, I was imagining the attribute is actually about inviting another author to a post. The specific post would then belong to both (all) authors, and thus be viewable and editable by them. But I also want to devise a system where all authors must mutually and more-or-less simultaneously agree to publish a post.

rytmis:
Some early thoughts I've tried to put to words. A bit disjointed, and none of this is meant as strongly-held positions on anything. 😃 

# Claims

From the Letterbook API perspective, _claims_ are key-value pairs attached to an _identity_ that the current request presents. ASP.NET Core presents claims-based identities as a ClaimsPrincipal object attached to the HttpContext. The principal has 1..* identities which have 0..* claims (although typically at least a Name claim). Even an unauthenticated user has a generic identity.

A principal can have many identities. The identities can be sourced from various providers, and the application itself may choose to augment the claims via a _claims transformation_. This means that whatever authentication scheme is used, it doesn't have to be the single source of truth for the claims.

# JWTs

Any third-party clients will have to use some sort of token-based authentication, so JWT it is because that's what we do these days. The example claims in the document, to me, look like permissions to perform a type of action, a typical example of a JWT _scope_ claim.

I tend to lean towards @mattly 's line of thinking in that permission/scope claims should rarely if ever be relational. The claims would more correctly answer the question "which _types_ of actions is this application permitted to perform on my behalf."

Since the process of logging in can involve an interactive part that allows the user to make choices, one of the choices could be "which of the things I have control over do I want to expose to this application." So the JWT _could_ contain claims representing all of the objects, if not the actual relations to those objects. Example: a token has Post:Draft, Post:Publish, Post:Read etc. and accounts `katherine_janeway` and `seven9`. This doesn't guarantee that the calling app can perform _any_ of the listed actions on _any_ of the listed accounts--it's a "necessary but not sufficient" condition.

Whether or not this is useful is really dependent on what we expect _third-party clients_ of the _first-party API_ to do: if we have a Mastodon-compatible API and Mastodon doesn't understand co-authoring, a Mastodon client won't really benefit from having a claim that allows access to a co-authored object? 

On the other hand, if we expect the client to use the first-party Letterbook API, we could just as well drop the objects from the claims and expose it as an API endpoint to be called every time the list of accessible accounts needs a refresh.

# ABAC

Authorization rules such as "can X perform Y _on object Z_" would then be decided by the API based on at least two factors:
- Does the current principal have permission to _perform this action_ (JWT)
- Does the actual user represented by this principal have permission to perform this action _on this particular object_ (secondary source, like database relation expressing co-author permissions)

# First-party client authentication scheme

If we ship a first-party web client that gets packaged and served along with the API, I'd argue that cookie authentication would make more sense than JWT. Cookies can be more secure, because with the HttpOnly flag, the browser sandbox won't allow it to leak. Tokens, on the other hand, have to be available to frontend code, so the frontend could exfiltrate them.

# Token revocation

We'll have to ship an OIDC token server of some sort if we want Mastodon client compatibility. Do we want to keep it architecturally separate from the API?

The way I see it, we don't expect other API resources to depend on our tokens for authentication. We'll have _clients_, sure, but the actual resource being invoked will always be _our API_. Which means we don't _have_ to separate the OIDC server from the API, and if we don't, we'll have direct access to its database. Which means that token expiration is not difficult or expensive, right?

jenniferplusplus:
This is good feedback, thanks. I'll have more thoughts later, but for now: there are some potentially high QPS and complex scenarios that I want to be able to support.

1. Sort of tag-team notification filtering. The idea is that the target of a twitter-style pile on could ask someone they trust to (help) monitor and screen their notifications. Assuming in the future we have near-real-time notifications, that could be a lot of interactions with delegated permission to authorize.
2. Actual API clients. I don't expect 3rd party UI apps to consume the 1st party API any time soon, but I think bots could. I want to be able to delegate granular permissions to bots (or really, to API keys), and support the kind of high volume interactions a bot might perform.

So, those are the really concrete thoughts that informed how I was approaching the claim structure.

I agree bringing multiple accounts into it is a bad idea. I had a similar notion as above about moderator access, but this is definitely the wrong tool for that job.

jenniferplusplus:
@rytmis re: OIDC and revocation.

Broadly yes, although I'm not entirely sure mastodon incorporates OIDC into their oauth2 flow. I do aspire to make Letterbook an OIDC provider either way, and we would just consume our own OIDC issued tokens. As with most everything, I would like to be able to host that in a unified package, or on its own. I actually imagine being a general purpose OIDC identity provider eventually. But that's a very long term concern.

In any case, I expect we would use paired refresh and authorization tokens. Revoking authz tokens would always be expensive. Refresh tokens are cheaper, since they're checked infrequently. Refresh tokens are what would correspond to a session and allow terminating those sessions.

jenniferplusplus:
Actually, I guess I also wanted to address using cookie tokens, although that's a little beyond the scope of this ADR. Basically, I don't think that cookies are intrinsically more secure than bearer tokens. It's a question of the threat model. And in my mind, malicious websites that try to perform XSRF attacks are a more pertinent threat than malicious clients that exfiltrate bearer tokens. Especially since malicious clients don't even need to exfiltrate tokens in order act maliciously.

PR #184, opened 2024-03-25
Enhancement/post api

### LINKED ISSUE #176: Posts API
This will be essentially the first proper letterbook api. So, I'll need to put some thought into api patterns.

Also, to keep the scope manageable, this is just about managing posts and threads. Not searching or listing or etc.

jenniferplusplus:
Adds the first of the the fist-party APIs: posting

Sorry it's huge, it's just that everything continues to be complete green space. Anyway, these are the newly added endpoints.

image

## Details
- Establishes some api patterns
- Scaffolds out some authorization interfaces
  - Probably not strictly necessary, but it helped with reasoning through the points where Authz decisions would happen, and the data needed, and I was having a ton of trouble handwaving that
- Sets up a lot of DTOs, fakers, mocks, etc
- Mappers/model binding for UUIDv7
  - This makes everything quite a lot easier to work with

## Related
- closes #176


Letterbook.Api.Tests/MapperTests.cs
jenniferplusplus, line 1: FYI @rytmis, since you'll likely be updating mappers

Letterbook.Api/Program.cs
```diff
@@ -132,60 +137,27 @@ public static void Main(string[] args)
...
-                        Scheme = "oauth2",
-                        Name = "Bearer",
-                        In = ParameterLocation.Header,
-                    },
-                    new List<string>()
-                }
-            });
-            options.OperationFilter<RequiredHeaders>();
-        });
+        builder.Services.ConfigureSwagger();
```
jenniferplusplus, line 146: This all moved into the extension method in `Letterbook.Api`

Letterbook.Core.Tests/Fixtures/IntegrationTestExtensions.cs
jenniferplusplus, line 1: This is extracted from `Letterbook.Adapter.Db.IntegrationTests`, in order to make it available as consistent seed data for all integration tests

PR #185, opened 2024-03-28
Configure RelationalContext ahead of time

jenniferplusplus:
This is a general architectural tweak that I've been meaning to do for a while. Then EFCore helpfully threw errors at me to let me know it was important for performance, too. So I made it more urgent.

This saves EFCore from having to create large numbers of DI containers, which can impact performance

## Details
- All of the work to configure db connections that used to happen in `OnModelCreating` is now done in DI
- This means `DbContextOptions` is already created, and `RelationalContext` doesn't need any injected dependencies
- That means EFCore can do a lot fewer service lookups and allocations, which is generally a good thing


Letterbook.Adapter.Db/RelationalContext.cs
```diff
@@ -32,29 +31,19 @@ public class RelationalContext : DbContext
...
     {
-        _config = config.Value;
-    }
-
-
-    protected override void OnConfiguring(DbContextOptionsBuilder options)
-    {
-        var builder = new NpgsqlDataSourceBuilder(_config.GetConnectionString());
-        builder.EnableDynamicJson();
-        options.UseNpgsql(builder.Build());
```
jenniferplusplus, line 49: This is the really critical bit. We no longer inject `IOptions<DbOptions>` to the `RelationalContext`, which winds up having significant performance impact under heavy load

Letterbook.Adapter.Db/DependencyInjection.cs
```diff
@@ -0,0 +1,25 @@
...
+
+public static class DependencyInjection
+{
+	public static IServiceCollection AddDbAdapter(this IServiceCollection services, IConfigurationSection config)
+	{
+
+		var dbOptions = config.Get<DbOptions>() ?? throw new ArgumentNullException(DbOptions.ConfigKey);
+
+		return services.AddDbContext<RelationalContext>(options => options.UseNpgsql(DataSource(dbOptions)));
+	}
```
jenniferplusplus, line 16: And we accomplish that by configuring the `DbContextOptionsBuilder` as part of calling `AddDbContext` in DI.

Issue #118, opened 2023-11-05
Document Http Signatures


jenniferplusplus:
Background on how and why they work, and our usage and requirements.

Notes from #110

The way http signatures works, you have to define a specification for how to build an input document, and then the input document is what's actually signed. The general reason for that is because HTTP is messy, and there's multiple ways to express most things. Like Host header + path vs using a fully qualified uri. Or multiple values in a header vs multiple instances of the same header. Different user agents will construct http messages in different ways. And that's before the kinds of changes that http intermediates can make to a message before it reaches its destination. So most of the spec is about how to construct this input document.

This bit is doing that reconstruction for received messages. If we build the document in a way that's different from the sender, then it won't verify. So, the risk is being overly restrictive, rather than overly permissive. And I'm okay with that.

The other consideration is that I want to (in the future) support multiple signatures so that we can use better/faster algorithms and up to date libraries implementing the signature spec. So, over time the mastodon verifier becomes a special case and could theoretically try to verify signatures that it just doesn't understand.

_Originally posted by @jenniferplusplus in https://github.com/Letterbook/Letterbook/pull/110#discussion_r1382431299_

strand:
Where should this documentation live? `docs/http-signatures.md` or somewhere else?

jenniferplusplus:
Hi!

And, uh, sure, that would be fine. Our https-sigs are still in kind of rough shape, but if you want to get something written up, I would love that.

jenniferplusplus:
Updating this to mark as not ready to start. Our http signatures are just like everyone else's right now. The really valuable thing to document will be how we can move to conform with the RFC, without breaking compat.

Which is not actually done, so documenting it is premature.

PR #186, opened 2024-03-29
HTTP Signatures documentation

johana-star:
This PR creates an HTTP Signatures doc. However, since it is based off of test examples and I lack project familiarity, it is almost certainly incorrect. 

## Details
Regarding probable incorrectness of this doc, I am fairly certain that the Inbox and Outbox are nested under the Actor in this data model, however, I have inbox and outbox examples under root, as this corresponds to some test data. Generally, I am treating this doc as a stub to begin with... Reviewers: are there listed paths which do not exist in Letterbook? Are there unlisted paths that are vital to understanding Letterbook's functioning? Does the order of this document generally make sense? What commentary is important to include around http signatures? Does having methods plus paths make sense when documenting Letterbook's HTTP Signatures? Are there other methods (POST/PUT/DELETE/etc) which are being used by Letterbook but not currently documented?

## Related
Addresses #118.


jenniferplusplus:
Thanks for starting this! I think there's probably some missing context, so let me back up.

HTTP signatures are used to provide a kind of federated authentication mechanism. Essentially, they make it verifiable that when we receive a POST to an activitypub endpoint, that it was actually sent by the actor that the POST body claims to be. So, it's only relevant when interacting with federated peers. For our own client APIs, we'll just use conventional bearer tokens. The endpoints where this is relevant are the inboxes, content urls, and other collections referenced under profiles. This is a not-comprehensive list. These will all eventually be handled by controllers under the ActivityPub folder in the API project. For now I think we just have Actor.

- `/actor/{id}`
- `/actor/{id}/inbox`
- `/actor/{id}/followers` (etc for other collections)
- `/<type>/{id}` (where `<type>` is like `note`, `image`, etc; ActivityPub types)
- `/<type>/{id}/likes`
- `/<type>/{id}/replies` (etc for other collections)

That said, the thing to document about our HTTP Signatures is less where they're used, and more how they're used. It's a common stumbling block for fedi devs. They're defined in [RFC 9421](https://www.rfc-editor.org/rfc/rfc9421.html). It might help to have that handy, but it's also important to know that Mastodon implemented an early draft of that spec from like 2017 and then just never updated. So everyone else is stuck working to that standard. Which sucks. But the basic scenario it's trying to solve is still the same. Specifically:

1. HTTP Signatures verify the integrity of (part of) an HTTP message
2. They do this by defining a set of properties from the message (like specific headers or url components) to sign over
    1. That definition is important, see number 5
4. They compose those properties into an ephemeral document, and sign that
5. Attach the signature and the definition, but not the ephemeral document, as new headers
6. The recipient must then use that definition to reconstruct the ephemeral document, and then verify the signature against that document

Which is convoluted. But it's largely necessary, in order to perform that message integrity function. It also permits unknown clients to authenticate themselves, protect their messages from tampering/spoofing, and make the authenticated messages un-replayable.

So, the kind of documentation that's most relevant is how we go about (re)composing and signing/verifying that ephemeral document, what fields we sign/require to be signed, how we lookup signing keys during verifications, the algorithms we can sign and verify, and so on. And at least half of that list is actually not implemented yet. It's probably also helpful to document how we would handle unsigned or unverifiable requests. Also not implemented.

jenniferplusplus:
The other thing I want to document is how we can try to move forward to support the actual spec from RFC 9421, instead of staying stuck on the 7+ year old draft. The library I built this on is quite robust, and can certainly allow us to work with 9421 signatures. So we would need to explore (and socialize among other projects) options to dual-sign for a while, and then eventually discontinue the 2017-draft style signatures.

jenniferplusplus:
I think mastodon targeted either the cavage-07 or cavage-08 draft. I don't remember, but they're very similar. Anyway, the whole history of the RFC is available here https://datatracker.ietf.org/doc/draft-cavage-http-signatures/07/

We have auto-generated OpenAPI docs, so these things are semi-self documenting. We do need to specify auth mechanisms explicitly, but the set of available endpoints is really easy to get. The default launch settings should just open a browser to the swagger ui when you run the app.

johana-star:
Regarding running the app... How do I do that locally? I didn't see anything in the README about starting the app, and I haven't worked with a C Sharp project before, so I think I'm missing some obvious step zero stuff.

While looking through the project I noticed `Endpoints.http`, which has several method and url pairs, is this related to the third party http signatures in some way?

Thank you for your valuable, insightful feedback. I've completely redrafted this document based on it. What would need to be revised or added for this PR to be accepted?

jenniferplusplus:
No, that endpoints file is a jetbrains http test client thing. It's also kind of a mess, and I should really clean it up. The contributing guide has some more detailed instructions to get the app started. The short version is you'll need to install the dotnet sdk, start up the db, migrate the db, then run then app.

https://github.com/Letterbook/Letterbook/blob/main/CONTRIBUTING.md#build-and-run

jenniferplusplus:
This is starting to overlap a lot with work that @rytmis has in progress. It might make more sense to change focus. Unless you'd both like to collaborate on it?

jenniferplusplus:
@johana-star We recently added an actual front-end, and there's a ton that needs to be built out, there. I thought that might be a more familiar and approachable space to work in. This is the first front end PR that isn't pure scaffolding, if you want to have a look #209

johana-star:
Closing as this 1) feels like a superficial approach which 2) I don't have the guidance or insight to make meaningful enough for merge.

Will review the contributing guide and see about getting a local environ running. The frontend PR you mentioned is closed now, man April has sped by. :)

PR #204, opened 2024-04-11
Enhancement/web

### LINKED ISSUE #196: Add Razor front-end harness
## What is the goal of this feature?

Setup the necessary foundation for front-end work

## Describe the solution you'd like

Per discussion on Discord some months back, we need a Blazor harness setup in the app to enable front-end work. Sam had mentioned they were leaning towards the Blazor Server model.

### Describe alternatives you've considered

We talked about a bunch of alternatives, mostly JS-centric single-page app models such as React, Vue, or Svelte, and don't believe those make sense. Blazor fits into the .NET world.

### Additional context

If you can get me to a "Hello World" page, I can start poking around from there. Given that #167 and authentication is still in the works, perhaps we can identify some good starting front-end features that don't require authentication, such as displaying a local timeline.

jenniferplusplus:
Add an initial web view

The dev workflow is essentially the same, except you should launch `Letterbook.Web` instead of `Letterbook.Api` if you want the web view. You still need Postgres, the docker compose still works, etc.

image

If you're inclined to review it, then you should focus on these changes. Everything else is unmodified new project scaffolding.
- Letterbook.Adapter.ActivityPub/*
- Letterbook.Api/*
- Letterbook.Web/Program.cs

## Details

- [ ] Currently this is just a little refactoring in the DI config, plus a basic razor project scaffold
- [x] I still need to update the docs
- [x] I would like to have at least one interesting page, as a sample to work from
- [ ] In the long term, we want to be able to run the web and api separately
- [x] All the services should be setup in DI and available to use in Pages
- [ ] Oh, right, there are no tests. We need to figure that out, too.

## Related
- Closes #196


Letterbook.Web/Program.cs
```diff
@@ -0,0 +1,84 @@
...
+		app.UseHealthChecks("/healthz");
+		app.MapPrometheusScrapingEndpoint();
+		app.UseWebFingerScoped();
+
+		app.UseAuthentication();
+		app.UseAuthorization();
+
+		app.UseSerilogRequestLogging();
+
+		app.MapRazorPages();
```
runewake2, line 78: Did we intend to use Blazor for the frontend or Razor pages? This looks like a Razor webapp.

Letterbook.Web/Program.cs
```diff
@@ -0,0 +1,84 @@
...
+		app.UsePathBase(new PathString("/api/v1"));
+		app.MapControllers();
```
runewake2, line 80: How do we intend to connect the frontend and backend together?

Do we expect this server will proxy calls to a separately hosted Letterbook.API server?

Letterbook.Web/Program.cs
jenniferplusplus: We have discussed blazor in the past, but I think it works against some of the project's core goals. Razor seems more approachable for people with a general web background, and is likely simpler to scale. I summarized that thinking here https://github.com/Letterbook/Letterbook/issues/196#issuecomment-2046015222

Letterbook.Web/Program.cs
jenniferplusplus: The web project actually just hosts the api endpoints, alongside the razor pages. That's what `MapControllers()` is doing.

I expect that some views can use built in razor forms and otherwise just call the API from javascript.

runewake2:
Regarding testing, how do you feel about using Playwright to do that? https://playwright.dev/

jenniferplusplus:
Yeah that's probably the best option for UI functional tests.

jenniferplusplus:
All of the bootstrap and jq is scaffolding for a new razor project. I don't normally do front end work in dotnet, so I don't have a sense for what's normal in that space. I'm certainly not attached to either framework. What I really want at this stage is the templating.

jenniferplusplus:
All of the bootstrap and jq is scaffolding for a new razor project. I don't normally do front end work in dotnet, so I don't have a sense for what's normal in that space. I'm certainly not attached to either framework. What I really want at this stage is the templating.

Letterbook.Web/Pages/Profile.cshtml
```diff
@@ -0,0 +1,21 @@
...
+    <title>@Model.Handle</title>
+</head>
+<body>
+<div class="text-center">
+    <h1 class="display-4">@Model.Handle Profile Page</h1>
+    <p>ID: @Model.Prof.GetId25()</p>
+</div>
+<code>@Model.Json</code>
+</body>
+</html>
```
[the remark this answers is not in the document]
jenniferplusplus: good catch, i removed that

PR #206, opened 2024-04-16
Add playwright test project

runewake2:
Creates a project for #205 and adds a quick Playwright test project.

Still needs GitHub action integration - this only runs locally right now.

Playwright isn't well supported in XUnit, This project is built on top of NUnit. We could probably get XUnit to work here, but it'd have a number of limitations and would require a descent amount of work to orchestrate the parts Playwright is doing for us automatically in NUnit (or MSTest).

Issue #205, opened 2024-04-13
Add Playwright E2E Testing for Frontend Site


runewake2:
# Purpose

We're introducing a Letterbook frontend built using Razor pages starting with #204. We need a way to test the functionality we're introducing in this frontend and validate its features as the project grows.

One of the ways we can do that is through the introduction of end to end (E2E) testing. Playwright is one of a few options for this with built-in support for .NET that can provide a way for us to build and run e2e tests against a variety of web browsers automatically as part of a PR's validation.

E2E testing is needed to ensure Letterbook functionality and behavior as exposed to users continues to operate in a stable, predictable and reliable way.

## Steps

- [ ] #206 Introduce a new Playwright based e2e test project
- [ ] Execute Playwright tests as part of PR's when a web or playwright test project change is detected in the PR (depending on how our dependency structure looks we may want to execute these tests regardless of the project that is changed)

## Related

- #196

jenniferplusplus:
I'm actually fine with running integration tests on every change. I'm a lot less protective of github action's time than I am of contributors.

PR #207, opened 2024-04-16
Add playwright testing project to main

runewake2:
Creates a project for https://github.com/Letterbook/Letterbook/issues/205 and adds a quick Playwright test project.

Still needs GitHub action integration - this only runs locally right now.

Playwright isn't well supported in XUnit, This project is built on top of NUnit. We could probably get XUnit to work here, but it'd have a number of limitations and would require a descent amount of work to orchestrate the parts Playwright is doing for us automatically in NUnit (or MSTest).

This is a retargeting of #206.


jenniferplusplus:
I take it that it's difficult to get a running instance of letterbook that playwright can target? I was looking around a little, and [this guide](https://khalidabuhakmeh.com/end-to-end-test-with-aspnet-core-xunit-and-playwright) seems like a good way to start the web server in the tests. The most relevant bit is here

```csharp
public class WebServerFixture : IAsyncLifetime, IDisposable
{
    private readonly IHost host;
    private IPlaywright Playwright { get; set; }
    public IBrowser Browser { get; private set; }
    public string BaseUrl { get; } = $"https://localhost:{GetRandomUnusedPort()}";

    public WebServerFixture()
    {
        host = Program
            .CreateHostBuilder(null)
            .ConfigureWebHostDefaults(webBuilder => {
                webBuilder.UseStartup<Startup>();
                webBuilder.UseUrls(BaseUrl);
                // optional to set path to static file assets
                // webBuilder.UseContentRoot();
            })
            .ConfigureServices(configure => {
                // override any services
            })
            .Build();
    }
```

The other option that seemed promising was the [background action for GHA](https://github.com/marketplace/actions/background-action). We could presumably use that to run an instance of the server and then run playwright tests in a subsequent step.

runewake2:
Yeah, I think that could work, sort of searching for some relatively performant way to run the web project in the background during tests. If we don't care about the playwright limitations in xunit Khalid's xunit example in that article is also a lot less messy than other examples I've seen - maybe worth a shot. I can take a look this evening I think.

ben-biddington:
Can I help with this?

Out of interest, is the plan to keep the UI source in this repository?

jenniferplusplus:
@runewake2 have you had a chance to come back to this? If not, can ben-piddington pick it up?

Re where the UI source will live, yes, it will stay in this repo. I think it will always be true that the letterbook application itself will all be located together in one repo/solution. It's conceivable that in the future we might move some supporting bits out to independent packages (http signatures come to mind, and the SSG for our docs).

But in any case, I think splitting things up at this point would add more overhead than it saves.

ben-biddington:
Actually I can't help, my computer is too slow to be practical. For what it's worth though, I reckon `WebServerFixture` is worth a try

runewake2:
Yes, that'd be great! I'm a bit behind with other work and haven't been able to loop back to this. Only the github action work should be left to get this running and some updates to catch up version wise. If there are any questions or issues happy to jump in as I can.

ben-biddington:
Righto, will have a look tomorrow 👍🏼

ben-biddington:
Okay this build is now passing with one test that is using an on-demand server [as described](https://khalidabuhakmeh.com/end-to-end-test-with-aspnet-core-xunit-and-playwright):

https://github.com/Letterbook/Letterbook/pull/320

The build does now essentially follow the startup guide from CONTRIBUTING.md, not sure if that is okay or not.

jenniferplusplus:
Added in #320

PR #209, opened 2024-04-16
basic profile display

mattly:
This displays basic profile information on the profile page. I edited my profile in postgres directly to provide description and custom fields: 
image

## Details
- I really do not want to expose the main `Prof` model to the template for reasons discussed in Discord. For posterity, the concern is about keeping an inventory of work done (database access, etc) outside the templates. The main model has a number of methods available that can do all sorts of things, and those should not be available to the template. I'm still new to dotnet / csharp, but the way I found to do this was by making the `Prof` field `private protected` and nullable; anything else I tried either still exposed that field to the template or gave me compiler warnings/errors.

## Related
Further things that could be useful here:
- how do I get a list of posts created by the user? I've looked through post the Posts and Timeline services, but neither has what looks like an appropriate method currently. My guess is this isn't implemented yet?
- how to get the server's url, to fill out the handle? right now it displays `@admin` but it should display `@admin@example.com` or whatever the host is.

### DIFF
--- a/Letterbook.Web/Pages/Profile.cshtml
+++ b/Letterbook.Web/Pages/Profile.cshtml
@@ -4,10 +4,32 @@
 @{
     ViewData["Title"] = Model.Handle;
 }
-<div class="text-center">
-    <h1 class="display-4">@Model.Handle Profile Page</h1>
+
+<div>
+    <h1>
+        <span>@Model.DisplayName</span>
+        <span>@@@Model.Handle</span>
+    </h1>
+    
+    <div>
+        <div>@Html.Raw(@Model.Description)</div>
+        
+        <dl>
+            @foreach (var field in @Model.CustomFields) {
+                <dt>@field.Label</dt>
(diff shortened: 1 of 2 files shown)


jenniferplusplus (approved):
> how do I get a list of posts created by the user? I've looked through post the Posts and Timeline services, but neither has what looks like an appropriate method currently. My guess is this isn't implemented yet?

Correct. The straightforward way would be to just add a Posts property on Profile, which I'm sure would work fine. The relation already exists in the schema. I just haven't done it.

> how to get the server's url, to fill out the handle? right now it displays @admin but it should display @admin@example.com or whatever the host is.

You can get it from the CoreOptions. You'll need to add an `IOptions<CoreOptions>` constructor arg, and then unwrap the Option to get the value. The api controllers do the same thing, so you can use that as an example.

Anyway, I'm happy with this. If you want to work on the couple of things you called out, go for it. Otherwise, I think this is good to merge.

PR #211, opened 2024-04-21
Reduce integration test conflicts

jenniferplusplus:
Sets up (some, more to come) integration tests to run with a dedicated database per class, which eliminates the biggest source of conflicts and makes them more parallelizable.

## Details
- Override the config used in tests
- Set a unique database name, based on the test class
- Skip the bootstrap logger during tests

### DIFF
--- a/Letterbook.Api.IntegrationTests/Fixtures/HostFixture.cs
+++ b/Letterbook.Api.IntegrationTests/Fixtures/HostFixture.cs
@@ -1,53 +1,99 @@
+using System.Text.Json;
 using Letterbook.Adapter.Db;
 using Letterbook.Core;
 using Letterbook.Core.Models;
+using Letterbook.Core.Tests;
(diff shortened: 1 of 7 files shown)


Letterbook.Api.IntegrationTests/Fixtures/HostFixture.cs
```diff
@@ -1,53 +1,99 @@
...
+		builder
+			.UseConfiguration(cfg)
+			.ConfigureServices(services =>
+			{
+				// SeedAdminWorker executes before we have a chance to create the test database
+				// So we just remove it
+				var seedDescriptor = services.SingleOrDefault(d => d.ImplementationType == typeof(WorkerScope<SeedAdminWorker>));
+
+				services.Remove(seedDescriptor);
+			});
```
jenniferplusplus, line 95: This method is doing almost all of the work in this change

PR #213, opened 2024-04-22
ADR-10 Razor Pages

jenniferplusplus:
Adding an ADR to note the Razor web frontend

## Details
- [Link to the doc preview](https://github.com/Letterbook/Letterbook/pull/213/files?short_path=53119c0#diff-53119c0c0ee0124778e3f64e0d737674c548a471566bf1d03c6b0b315d8c5d8a)

## Related
- re: #204


jenniferplusplus:
It's probably a good to do something, but I'm not sure I want dates. The thing I mostly had in mind with naming was making it easy to reference them later. So when new contributors ask, we can say easily that we discussed the state of the software architecture in adr-08, for example.

I also imagine including them in the docs site, and for that to provide a more intentional experience than just markdown files in github. Like adding some front matter metadata, and organizing them based on that, rather than just lexicographically sorting them by filename.

PR #218, opened 2024-04-28
enhancement(web): default register, login, logout

jenniferplusplus:
Adds default(ish) account registration, login, and logout pages from Aspnet Identity, and (actually) default account maintenance pages, to allow managing account data like change email and password resets.

The default UI is not one that we'll want to keep long term, but I think it will be helpful as a starting point, and just during development.

image

## Details
- Use identity default UI via `AddDefaultUI()` builder method
- Override a handful of basic pages so they'll actually work
  - Scaffolding the pages is basically impossible in app where the dbcontext is in another project than the mvc app.
  - Many of the unscaffolded defaults don't actually work


Letterbook.Api/DependencyInjectionExtensions.cs
```diff
@@ -85,6 +85,16 @@ public static void UseWebFingerScoped(this IApplicationBuilder app)
 			}));
 	}
 
+	public static IdentityBuilder AddIdentity(this IServiceCollection services)
+	{
+		return services.AddIdentity<Account, IdentityRole<Guid>>(options =>
+			{
+
```
jenniferplusplus, line 92: This is where we would configure things like password policies

jenniferplusplus:
Moving to a new branch, for a less confusing commit history #219

Issue #191, opened 2024-04-01
Posts - Outbound federation


jenniferplusplus:
We need to send Create, Update, and Delete activities related to our own posts. This is mainly powered by the events that are already emitted by the core PostService.

- [x] build mappers for Post -> ASObjects
- [x] map the Post to an ASObject[^2] (we only have Notes at the moment, so it's okay to just handle that case) and wrap it in an ActivityPubDocument[^1].
- [x] use a task worker to handle and consume post events:
  - [x] Published (might need to revisit the published event, and only emit the first time it happens)
  - [x] Updated (maybe just published + check create/update/publish dates?)
  - [x] Deleted
  - [x] Like (bonus!)
  - [x] Share (bonus!)
- [x] lookup the inboxes[^3] (shared inbox when possible) for all of the members of the post's audience, and schedule them for delivery through the ActivityMessage service.
- [ ] test interop with mastodon, and ideally other peers
  - [x] text content is legible
  - [ ] ~~mentions are legible~~ (skipping)
  - [ ] ~~mentions trigger notifications~~ (skipping)

Update:

Task workers don't look like they used to. The high level flow is the same, but now they're all implemented in the Letterbook.Workers project, as a `MassTransit.IConsumer<TContract>` class. Outbound federation would consume the same messages as `TimelineConsumer`

[^1]: See the ActivityMessagePublisher for an example of building such a document
[^2]: I don't think these mappers are built yet. Please reach out if you have any question at all
[^3]: These queries definitely aren't built yet. Let me know if you're taking on this issues, and I'll fix that

jenniferplusplus:
This might be an interesting thing for people work on. Assigning myself as a reminder to check and see what's missing in order to make this ready to start.

PR #224, opened 2024-05-04
Internals/event refactor

jenniferplusplus:
Cleans up and refactors observable event handling.

These events facilitate doing deferred work outside the HTTP request/response cycle. They're presently backed by in-memory Rx.Net Subjects, but they could also be handled by a message broker in higher-capacity deployments. This supports loose coupling between side-effects, future scale out, and enables the potentially long running work and half-duplex communication required by ActivityPub.

## Details
- Introduce a generic MessageHost and MessageObserver
  - This allows new message types to be observed and processed by implementing only the necessary MessageWorker class
  - Refactor DeliveryWorker to utilize the new generics
- Start producing PostEvents
  - Note there is no observer for these events, yet
- Begin introducing explicit types for message Data fields
- Refactor message channels to key on the channel type, rather than the model class
  - This gives somewhat better type safety
- Removed an old service for event types that we're not actually going to use

## Related
- Prep for #191


Letterbook.Core/EventServiceBase.cs
```diff
@@ -0,0 +1,21 @@
...
+/// <typeparam name="T"></typeparam>
+public abstract class EventServiceBase<T> where T : IEventChannel
+{
+	protected IMessageBusAdapter _messageBusAdapter;
+	protected IObserver<CloudEvent> _channel;
+
+	protected EventServiceBase(IMessageBusAdapter messageBusAdapter)
+	{
+		_messageBusAdapter = messageBusAdapter;
+		_channel = messageBusAdapter.OpenChannel<T>(GetType().Name);
```
jenniferplusplus, line 19: Note the EventServiceBase is generic, and uses the type property as the channel key

Letterbook.Core/PostEventService.cs
jenniferplusplus, line 1: This is the only really genuinely new thing

jenniferplusplus (commented):
There's a lot of relatively straightforward rename refactoring included. This is mainly to clarify what the relationship is between various event-related types. Also, somewhat better organization.

I called out the more interesting changes below.

PR #225, opened 2024-05-11
Enhancement/timeline models

### LINKED ISSUE #222: Update Timeline models & entities
Timelines shouldn't need hand crafted SQL and weird isolated models anymore. They should be able to just use Posts. That was a lot of the point of ADR-07, after all

jenniferplusplus:
Updates Timeline models and related services and adapters. Timelines are now streams of `Post`, which has enough content and addressing data to build partial views of a collection of posts

## Details
- IFeedsAdapter now exposes some `IQueryable`s. Taking proper advantage of this requires core services to import EFCore, but I'm starting to be of the opinion that this is okay. The query object pattern is _much_ easier to work with than the kind of quasi-repositories we have now, and that's what `IQueryable` does.
  - Feedback welcome
- Basically just throws away the old timeline stuff. It was some of the earliest code written in the project. It helped to overcome the blank page and work out some data modelling questions, and now it needs to go.
- Trying to get better about adding xmldocs to interface methods. At some point, I want to make them mandatory.

The test data attempts to generate a distribution of posts to profiles that is at least semi-realistic. There should be a small number of very active posters, and a long tail of other profiles that post considerably less. Also, the substantial majority of posts should be replies. This doesn't handle shared posts, but in the future, that should likely be the largest category. The distribution should look _more or less_ like this

post distribution

## Related
- closes #222


Letterbook.Adapter.TimescaleFeeds/EntityConfig/ConfigureTimelinePost.cs
```diff
@@ -0,0 +1,33 @@
...
+{
+	public void Configure(EntityTypeBuilder<TimelinePost> builder)
+	{
+		// This key is purely for EFCore's benefit, we'll likely never (need to) query on it
+		// TODO(profiling): Evaluate the performance impact of this key, and compare with the hastle of using `ExecuteSqlAsync` to do
+		// custom INSERTs
+		// If ExecuteInsert() ever happens, use that instead and drop this index
+		// https://github.com/dotnet/efcore/issues/29897
+		// Also, explore using BulkExtensions https://www.nuget.org/packages/EFCore.BulkExtensions
+		builder.HasKey(model => new { model.Time, model.PostId, model.AudienceId });
```
warriordog, line 17: if we don't use this key directly, then should we just use an auto-increment INT key or something? That seems like it will be more efficient than a compound key.

Letterbook.Adapter.TimescaleFeeds/EntityModels/TimelinePost.cs
```diff
@@ -0,0 +1,100 @@
...
+	/// <summary>ID of the Thread to which the Post belongs</summary>
+	public required Guid ThreadId { get; set; }
+
+	public static implicit operator Post(TimelinePost p) => new Post
+	{
+		Id = p.PostId,
+		Thread = new ThreadContext()
+		{
+			Id = p.ThreadId,
+			FediId = default!,
```
warriordog, line 65: is it safe to forcibly set this non-nullable field to null?

Letterbook.Adapter.TimescaleFeeds/EntityModels/TimelineProfile.cs
```diff
@@ -0,0 +1,30 @@
+﻿using Letterbook.Core.Models;
+using Medo;
+
+namespace Letterbook.Adapter.TimescaleFeeds.EntityModels;
+
+public class TimelineProfile
+{
+	// nameof(Profile.Id), nameof(Profile.FediId), nameof(Profile.DisplayName), nameof(Profile.Authority)
```
warriordog, line 8: commented out code

Letterbook.Adapter.TimescaleFeeds/FeedsAdapter.cs
```diff
@@ -131,22 +41,26 @@ public Task<int> RemoveFromTimelines(Models.Post post)
...
-			.Where(t => keys.Contains(t.AudienceKey))
-			.Where(t => includeBoosts || t.BoostedBy != null)
-			.GroupBy(t => t.EntityId).Select(g => g.First())
-			// Equivalent to DistinctBy, but EFCore won't translate that for some reason
-			// .DistinctBy(t => t.EntityId)
-			.Take(limit);
+		var keys = audiences.Select(a => a.FediId);
+		return _feedsContext.Timelines
+				.AsNoTracking()
+				.OrderBy(t => t.Time)
```
warriordog, line 50: is EF Core smart enough to re-order this query? In older versions (framework era), this would prevent the query from translating correctly. The following "where" clauses would run in application code instead of SQL.

Letterbook.Adapter.TimescaleFeeds/FeedsAdapter.cs
```diff
@@ -162,14 +76,15 @@ public Task Cancel()
 
 		return Task.CompletedTask;
 	}
+
 	public Task Commit()
```
warriordog, line 80: suggestion: make this async to preserve the stack trace

Letterbook.Adapter.TimescaleFeeds/FeedsContextFactory.cs
```diff
@@ -4,14 +4,25 @@
...
 
 /// <summary>
-/// This class is only used by EFCore design tools, to generate migrations 
+/// This class is only used by EFCore design tools, to generate migrations
 /// </summary>
 public class FeedsContextFactory : IDesignTimeDbContextFactory<FeedsContext>
 {
 	public FeedsContext CreateDbContext(string[] args)
 	{
+		var feedsDbOptions = new FeedsDbOptions
```
warriordog, line 13: should these be loaded from config?

warriordog (approved):
looks great! Left a few questions and suggestions, but nothing major.

Letterbook.Adapter.TimescaleFeeds/EntityConfig/ConfigureTimelinePost.cs
jenniferplusplus: I expect this to be a very insert heavy table, and I worry about using db-generated keys, because I think that's more likely to produce conflicts/add load.

The best thing might be a Guid key that we just never use.

Letterbook.Adapter.TimescaleFeeds/EntityModels/TimelineProfile.cs
jenniferplusplus:
```suggestion
```

Letterbook.Adapter.TimescaleFeeds/EntityModels/TimelinePost.cs
jenniferplusplus: Yeah. Safe-ish, anyway. The only instance where this constructor should be used is when we already have a thread ID, and we just need to look it up to fill out the rest of the object.

It would be a bug for an object created this way to be written to the db, and the query would fail. So, it would produce some error messages and some action would fail, but I don't think it could corrupt any data.

PR #226, opened 2024-05-14
Internals/entry points

jenniferplusplus:
This creates a proper universal host project, to serve as the single service entry point for simple configurations. It also sets up the other host projects as specialized services. It also adds a new Workers host project, based on MassTransit, which will manage the various workers we require.

> [!Warning]
> After this update, the recommended way to get started as a developer is to launch the `Letterbook` project

## Details
- Projects that host a service
  - `Letterbook` - All-in-one, simple config and deployment. Includes all of the below.
  - `Letterbook.Api` - Just the API (first party, mastodon, and activitypub)
  - `Letterbook.Web` - Just the Web UI
  - `Letterbook.Workers` - Just the message bus workers

## Related
- Prep/yak shave for #222


jenniferplusplus:
Splitting out host project refactor from MassTransit workers

See #227

Issue #143, opened 2023-12-28
[Epic] Basic timelines


jenniferplusplus:
The counterpart to posting is reading other people's posts. Timelines are how you do that. We need to build them. Unfortunately, I'm not sure how to structure a lot of this. If you have thoughts, we should talk though it. Either way, this is something we should plan to iterate on.

## To do
- [x] #222
- [x] #203
  - [x] need to be able to lookup timeline entries on dynamically sliding time windows
  - [x] first need to look up appropriate audiences to query for
- [x] #228
- [x] #232
- [x] #248

> [!TIP]
> Update the above list as necessary. Convert to issues when ready.

## Goal
This is a little bit exploratory, so the first goal is really to get a better idea of what the correct design and behavior even are. We definitely want to populate a timeline with *something*, and we can iterate from there. So really, get to some concrete thing, because that's easier to talk about. Then we'll talk some more.

PR #230, opened 2024-05-20
Internals/mass transit workers

jenniferplusplus:
Replaces the improvised in-memory queue and workers with MassTransit.

(Sorry for the big PR. It hits a lot of files, but quite a lot of those are simple renames)

## Details
- Removes RxMessageBus
- Adds MassTransit
- Updates publishers to use message bus and types contracts (yay!)
- Updates delivery worker to use MassTransit

## Related
- More prep and/or yak shaving for #143
- And #228


Letterbook.Workers/Consumers/DeliveryWorker.cs
```diff
@@ -0,0 +1,38 @@
+using Letterbook.Core;
+using Letterbook.Core.Adapters;
+using Letterbook.Core.Models;
+using Letterbook.Workers.Contracts;
+using MassTransit;
+
+namespace Letterbook.Workers.Consumers;
```
jenniferplusplus, line 7: Git decided this is a new file, rather than a rename, which is unfortunate. This and the other masstransit workers are just updates of workers that used to be in core.

Letterbook.Workers/Publishers/ActivityMessagePublisher.cs
```diff
@@ -0,0 +1,56 @@
...
+using ActivityPub.Types.Conversion;
+using Letterbook.Core;
+using Letterbook.Core.Adapters;
+using Letterbook.Core.Extensions;
+using Letterbook.Core.Models;
+using Letterbook.Workers.Contracts;
+using MassTransit;
+using Microsoft.Extensions.Options;
+
+namespace Letterbook.Workers.Publishers;
```
jenniferplusplus, line 11: Similar, the MassTransit publishers are updates of the event* services that were previously in core. They even still implement the same interfaces.

Letterbook.Core/Workers/EventObserver.cs
jenniferplusplus, line 1: The homebuilt observers are all just removed

Letterbook.Adapter.RxMessageBus/RxMessageBus.cs
jenniferplusplus, line 1: The RxMessageBus services are just removed, and replaced by MassTransit, which is much more robust and easy to work with

PR #238, opened 2024-05-28
Internals/integration test filter

jenniferplusplus:
Adds a solution filter that excludes integration test projects. This allows us to keep all of our test projects in a single solution, and to still opt to run only zero-dependency unit tests in CI

## Details
- thanks to @AnthonySteele for the proposal


AnthonySteele (approved):
Looks nice

Issue #214, opened 2024-04-22
Add Profile Management to UI


mattly:
## What is the goal of this feature?
Provide affordances for the user to update various aspects of their profile (derived from #141 )

## Describe the requirements & constraints
On Mastodon, these actions are on their own page, which has four separate tabs
1. Edit Profile
    - Display Name 
    - Description
    - Custom Fields
    - Profile Picture
    - Header Picture
    - This is an automated Account
2. Privacy and Reach
    - Feature profile and posts in discovery algorithms
    - Automatically accept new followers
    - Include public posts in search results
    - Include profile page in search engines
    - Show follows and followers on profile
    - Display from which app you sent a post
3. Verification
    - copyable `a` tag with verification href
    - your verified links
4. Featured Hashtags

Some questions:
- Profile/Header picture - presumably depends on media support?
- this is an automated account - this is I think the "account type" flag?
- Privacy and reach stuff – how much of this is relevant to a single-user instance? how much of it is federated? how much of it should be included in single-user preview?
- verification - something to support for single user preview or not?
- Bio/Description - a) do we want to impose a character limit?  b) the Mastodon API & export provides this data as sanitized HTML, but mostly that seems to be in support of hyperlinks for hashtags. Otherwise the edit field is plain text. To what extent do we care about this? Turning newlines into <p> tags?
- Mastodon has this all on a single page behind "save changes", *including* the media files. Clicking that updates all the profile information on that tab at once. Letterbook's ProfileService so far looks to have very atomic profile update mechanisms (InsertCustomField, etc), and UpdateProfile is not yet implemented. I could go either way on the UI being declarative ("here is the new profile state") or imperative (amenable to click-to-edit on the profile page but requires fancier javascript).

jenniferplusplus:
> - Profile/Header picture - presumably depends on media support?

Yeah

> - this is an automated account - this is I think the "account type" flag?

Yes, that was the idea

> - Privacy and reach stuff – how much of this is relevant to a single-user instance? how much of it is federated? how much of it should be included in single-user preview?

Not much. Very little of it federates. The 2 things that do are search indexing and posts being featurable vs just public. I think Mastodon's implementation of both is pretty shit, and working around that will be annoying.

> - verification - something to support for single user preview or not?

I'd say yes, if we can. But not a reason to hold back from calling it ready for use.

[...]

PR #249, opened 2024-05-31
Enhancement/ Edit Profile

mattly:
It's not pretty, it's fairly blunt, and validation is kinda janky, but it does a roundtrip to the database.

It's been a while since I've done this sort of thing via form posts instead of doing JS to a JSON API directly. 

## Details
- It seems there are a bunch of different ways to handle form property bindings in Razor pages, but this was the only one I could get to work. 
- Model validation is a thing that can, and probably should, happen via System Annotations on the actual model object, which could then be exposed to the templates directly. Ideally any validation requirements could then be shared between this view logic and the API.
- Managing Custom Fields will require the ability to set them all in one go via the service class. I can probably tackle that but I want to get something up first.
- I'd like to get tests setup, especially for the form post endpoints, but I have no idea where to begin with that.
- I'll come back to validation and CSS later

## Related
- working towards #214 

image

### DIFF
--- a/Letterbook.Web/Pages/Profile.cshtml.cs
+++ b/Letterbook.Web/Pages/Profile.cshtml.cs
@@ -34,6 +34,7 @@ public class Profile : PageModel
 		Handle = "";
 		DisplayName = "";
 		Description = new HtmlString("");
+		CustomFields = new Models.CustomField[0];
 	}
 
 	


jenniferplusplus:
This looks great so far! Do you want me to add a unit test project for you, to give you a starting point?

jenniferplusplus:
Just checking, no pressure, but were you waiting on me for anything?

PR #250, opened 2024-06-01
Internals/web unit tests

jenniferplusplus:
Sets up an initial unit test project and mocks for Letterbook.Web

## Details
- I'm not sure if this is the best way to test a razor page controller
- But, given that it's supposed to be a page model, I think asserting the properties are as expected is reasonable?
- Also set warnings as errors in Web and Web.Tests

### DIFF
--- a/Letterbook.Web.Tests/WithMockContext.cs
+++ b/Letterbook.Web.Tests/WithMockContext.cs
@@ -1,0 +1,38 @@
+﻿using System.Security.Claims;
+using System.Security.Principal;
+using Letterbook.Core.Tests;
+using Microsoft.AspNetCore.Http;
+using Microsoft.AspNetCore.Mvc;
+using Microsoft.AspNetCore.Mvc.Abstractions;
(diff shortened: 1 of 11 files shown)


Letterbook.Web.Tests/WithMockContext.cs
```diff
@@ -0,0 +1,38 @@
...
+
+		MockPrincipal = new Mock<IPrincipal>();
+		MockPrincipal.Setup(x => x.Identity).Returns(Identity);
+		MockPrincipal.Setup(x => x.IsInRole(It.IsAny<string>())).Returns(true);
+
+		MockHttpContext = new Mock<HttpContext>();
+		MockHttpContext.SetupGet(m => m.User).Returns(ClaimsPrincipal);
+
+		var actionContext = new ActionContext(MockHttpContext.Object, new RouteData(), new ActionDescriptor());
+		PageContext = new PageContext(actionContext);
```
jenniferplusplus, line 36: This may or may not be sufficiently mocked, I'm not sure. But it's good enough for simple cases, at least.

PR #266, opened 2024-06-12
Docs/year 1 blog

jenniferplusplus:
Content for a first blog post: a summary of what we've done in the first year of Letterbook

## Details
- Thanks some people by name
  - Please let me know if I used the wrong name or pronouns, happy to fix that
  - Also happy to not name you, if you'd rather I didn't
- This is intended to be hosted by the razor press site
  - It will likely need to be adjusted to satisfy razor press assumptions when that's ready
  - github doesn't embed the toots, but I expect razor press can

These are the linked toots
- https://tech.lgbt/@jenniferplusplus/110567090111682149
- https://hachyderm.io/@runewake2/110567203751528214


jenniferplusplus:
Added in #297

PR #271, opened 2024-06-27
Enhancement/authentication

### LINKED ISSUE #212: Add authentication/authorization on API endpoints

jenniferplusplus:
We now lookup and set useful claims about related profiles on login, and perform some very basic authorization on most endpoints (mostly just being authenticated, we have no other requirements yet).

## Details
- Restructure profile claims as an array of enums
- Populates profile claims via middleware
  - Sets all the ProfileClaims from an account's LinkedProfiles
  - Sets a default `activeProfile` claim, that can be used to determine the acting profile in situations like determining if the user already follows another profile or not
- Includes a very little bit of conditional rendering in the profile page (this needs to be expanded, but it's useful as a proof of concept)

## Related
- Closes #212 
- Opting to retrieve profile claims per-request, based on discussion in #183


jenniferplusplus:
still need to add authorization on API endpoints, and enable them to use the Identity cookie. Not sure how much work that will be, but the rest of it is ready to review

Letterbook.Web/Pages/Profile.cshtml
```diff
@@ -11,8 +11,11 @@
         <span>@Model.Handle</span>
     </h1>
     
-    <a href="/@@@Model.BareHandle/edit">Edit</a>
-    
+    @if (User.Identity?.IsAuthenticated == true)
+    {
+        <a asp-page="ProfileEdit" asp-route-handle="@Model.BareHandle">Edit</a>
+    }
+
```
jenniferplusplus, line 18: Note the conditional rendering, and the preferred way to construct links to other pages

jenniferplusplus:
Most web views depend on having an `activeProfile` claim to handle relational elements. For example, when viewing a profile page, we need to know if it's the user's own profile or someone else's, in order to provide details like whether they're following each other, and to render a follow vs unfollow button. In the API, this is provided in the path. That doesn't work in web views, because most of them can be accessed anonymously, and URLs for web pages can't embed information about the current user. Instead, the `activeProfile` claim provides that information.

We will select a default active profile if the user hasn't set one. But, we need a UI feature to allow them the select and persist one for themselves.

PR #320, opened 2024-08-08
f/add playwright testing project to main #205

ben-biddington:
## Green build with development server

The changes here have the build passing with this end-to-end test from #207 

https://github.com/ben-biddington/Letterbook/blob/f/add-playwright-testing-project-to-main-%23205/Letterbook.Web.Tests.E2E/LandingPageValidationTest.cs

So I think you're free to continue in this style if you wanted to while experimenting with the on-demand server idea.

## On-demand server [WIP]

There are also some examples showing how you could run the tests against a server started at runtime:

https://github.com/ben-biddington/Letterbook/blob/f/add-playwright-testing-project-to-main-%23205/Letterbook.Web.Tests.E2E/SelfStartedWebServerExampleTest.cs

This could be useful if you wanted to be able to change the application under test. For example you may wish to substitute some of the services for fakes.

These tests use [`WebServerFixture`](https://github.com/ben-biddington/Letterbook/blob/f/add-playwright-testing-project-to-main-%23205/Letterbook.Web.Tests.E2E/Support/WebServerFixture.cs) to start the server.

I had some trouble with files paths getting it to partially work, and I think there is something missing with the razor pages which is why [HomepageHasCorrectTitleAndLinksToAdminProfile](https://github.com/ben-biddington/Letterbook/blob/f/add-playwright-testing-project-to-main-%23205/Letterbook.Web.Tests.E2E/SelfStartedWebServerExampleTest.cs#L46) is skipped.

(Thanks to @runewake2 for explaining why nunit has been chosen.)

## Related
- https://github.com/Letterbook/Letterbook/pull/207

### DIFF
--- a/.github/workflows/pull-request.yml
+++ b/.github/workflows/pull-request.yml
@@ -64,7 +64,7 @@ jobs:
         run: pwsh Letterbook.Web.Tests/bin/Release/net8.0/playwright.ps1 install firefox
 
       - name: E2E Test
-        run: BROWSER=firefox dotnet test Letterbook.Web.Tests.E2E --no-build
+        run: BROWSER=firefox dotnet test Letterbook.Web.Tests.E2E --no-build --configuration Release
         
       - name: Save test crash report
         if: failure()
--- a/.github/workflows/pull-request.yml
+++ b/.github/workflows/pull-request.yml
@@ -4,7 +4,7 @@ on:
   workflow_dispatch:
   push:
     branches:
-      - 'main'
+      - "main"
   pull_request:
     types:
       - opened
@@ -12,53 +12,71 @@ on:
       - synchronize
       - ready_for_review
     branches:
-      - 'main'
+      - "main"
 
 jobs:
   build:
     runs-on: ubuntu-latest
     env:
       NUGET_PACKAGES: ${{ github.workspace }}/.nuget/packages
-      
+
     steps:
       - uses: actions/checkout@v4
-        
+
       - name: Setup .NET Core
         uses: actions/setup-dotnet@v4
         with:
           dotnet-version: 8.0.x
-      
+
       - name: Package Cache
         id: package-cache
         uses: actions/cache@v4
         with:
           path: ${{env.NUGET_PACKAGES}}
           key: "${{ runner.os }}-nuget-${{ hashFiles('Directory.Packages.props') }}"
-      
+
       - name: Restore Packages
         if: steps.package-cache.outputs.cache-hit != 'true'
         run: dotnet restore
-      
+
       - name: Restore Tools
         run: dotnet tool restore
-      
+
       - name: Build
         run: dotnet build Letterbook.sln --configuration Release
-      
+
       - name: Test
         run: dotnet test Letterbook.UnitTests.slnf --configuration Release --logger GitHubActions --blame-crash --no-build
-        
+
+      - name: Run docker-compose
+        uses: hoverkraft-tech/compose-action@v2.0.1
+
+      - name: Seed database
+        run:
+          dotnet ef database update --project Letterbook.Adapter.Db/Letterbook.Adapter.Db.csproj &&
+          dotnet ef database update --project Letterbook.Adapter.TimescaleFeeds/Letterbook.Adapter.TimescaleFeeds.csproj &&
+          dotnet user-secrets set "HostSecret" "$(openssl rand -base64 32)" --project Letterbook
+
+      - name: Start server
+        run: dotnet run --project Letterbook.Web &
+
+      - name: Install Playwright Browsers
+        run: pwsh Letterbook.Web.Tests/bin/Release/net8.0/playwright.ps1 install firefox
+
+      - name: E2E Test
+        run: BROWSER=firefox dotnet test Letterbook.Web.Tests.E2E --no-build --configuration Release
+
       - name: Save test crash report
         if: failure()
         uses: actions/upload-artifact@v3
         with:
           name: TestResults
-          path: '**/TestResults/'
-      
+          path: "**/TestResults/"
+
       - name: Check Relational Migrations
         run: dotnet ef migrations has-pending-model-changes
         working-directory: Letterbook.Adapter.Db/
-        
+
       - name: Check Feeds Migrations
         run: dotnet ef migrations has-pending-model-changes
         working-directory: Letterbook.Adapter.TimescaleFeeds/
@@ -70,7 +88,7 @@ jobs:
           dotnet run --AppTasks=prerender --BaseUrl "${{env.custom_domain}}"
         env:
           custom_domain: "letterbook.com"
-      
+
       - name: Check Generated Files
         run: |
           CHANGED=$(git status --porcelain -uno)
--- a/.github/workflows/pull-request.yml
+++ b/.github/workflows/pull-request.yml
@@ -64,10 +64,9 @@ jobs:
         run: pwsh Letterbook.Web.Tests/bin/Release/net8.0/playwright.ps1 install firefox
 
       - name: E2E Test
-        run: dotnet test Letterbook.Web.Tests.E2E --no-build
+        run: dotnet test Letterbook.Web.Tests.E2E --no-build --configuration Release
         env:
           BROWSER: firefox
-          DOTNET_ENVIRONMENT: Release
         
       - name: Save test crash report
         if: failure()
--- a/.github/workflows/pull-request.yml
+++ b/.github/workflows/pull-request.yml
@@ -64,7 +64,7 @@ jobs:
         run: pwsh Letterbook.Web.Tests/bin/Release/net8.0/playwright.ps1 install firefox
 
       - name: E2E Test
-        run: BROWSER=firefox dotnet test Letterbook.Web.Tests.E2E --no-build --configuration Release
+        run: BROWSER=firefox dotnet test Letterbook.Web.Tests.E2E --no-build
 
       - name: Save test crash report
         if: failure()
--- a/.github/workflows/pull-request.yml
+++ b/.github/workflows/pull-request.yml
@@ -47,26 +47,6 @@ jobs:
       
       - name: Test
         run: dotnet test Letterbook.UnitTests.slnf --configuration Release --logger GitHubActions --blame-crash --no-build
-
-      - name: Run docker-compose
-        uses: hoverkraft-tech/compose-action@v2.0.1
-
-      - name: Seed database
-        run:
-          dotnet ef database update --project Letterbook.Adapter.Db/Letterbook.Adapter.Db.csproj &&
-          dotnet ef database update --project Letterbook.Adapter.TimescaleFeeds/Letterbook.Adapter.TimescaleFeeds.csproj &&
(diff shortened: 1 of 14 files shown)


runewake2:
Thanks for doing some work on this! One note from my old PR #207: [Playwright has some limitations against xUnit](https://playwright.dev/dotnet/docs/test-runners#xunit-support) that prevent it from having native support (this might be part of why your hitting weirdness in your test fixtures and why Playwright doesn't have these libraries available out of the box) and also means that parallel e2e tests won't work. That was part of why we had broken this code out originally even though it wasn't ideal.

Unsure if this is something we still want to capture or not.

ben-biddington:
@runewake2 Usually you can get parallelism with just xunit? In terms of sharing browsers that should also be taken care of by xunit I'd've thought.

If you want different parallel settings then I guess you may be better of splitting to a new assembly.

ben-biddington:
So the main thing here is I can't get the configuration right so that razor pages are rendered, at least I think that's why this `HomepageHasCorrectTitleAndLinksToAdminProfile` fails -- it encounters an empty screen.

ben-biddington:
On the upside, though, the tests all pass if run against a real server by supplying `BaseUrl`:

```shell
BaseUrl=http://localhost:5127 dotnet test Letterbook.Web.Tests/Letterbook.Web.Tests.csproj --filter DisplayName~LandingPageValidationTest
```

Letterbook.Web.Tests/Letterbook.Web.Tests.csproj
```diff
@@ -33,4 +34,10 @@
       <ProjectReference Include="..\Letterbook.Web\Letterbook.Web.csproj" />
     </ItemGroup>
 
+    <ItemGroup>
+      <Reference Include="ServiceStack">
```
jenniferplusplus, line 38: do we actually need this for something? ServiceStack powers the SSG for the docs site, but I kind of hate it and I want to get rid of it eventually, not pull it into the main app.

runewake2:
The xUnit team to my knowledge strongly discourages this or recommends disabling parallelism in tests to prevent flakiness here, the Playwright team has made similar statements (both were linked in my last comment). From what I've read their recommendations seem like good ones.

.github/workflows/pull-request.yml
```diff
@@ -44,10 +44,28 @@ jobs:
...
         run: dotnet build Letterbook.sln --configuration Release
+
+      - name: Install Playwright Browsers
+        run: pwsh Letterbook.Web.Tests/bin/Release/net8.0/playwright.ps1 install
+
+      - name: Docker compose up
+        run: docker compose up -d
+
+      - name: Seed database
+        run: dotnet tool restore && dotnet ef database update --project Letterbook.Adapter.Db/Letterbook.Adapter.Db.csproj && dotnet ef database update --project Letterbook.Adapter.TimescaleFeeds/Letterbook.Adapter.TimescaleFeeds.csproj
```
jenniferplusplus, line 55:
```suggestion
        run: dotnet ef database update --project Letterbook.Adapter.Db/Letterbook.Adapter.Db.csproj && dotnet ef database update --project Letterbook.Adapter.TimescaleFeeds/Letterbook.Adapter.TimescaleFeeds.csproj
```

Tools should already be restored from a few steps ago

.github/workflows/pull-request.yml
```diff
@@ -44,10 +44,28 @@ jobs:
       
       - name: Build
         run: dotnet build Letterbook.sln --configuration Release
+
+      - name: Install Playwright Browsers
+        run: pwsh Letterbook.Web.Tests/bin/Release/net8.0/playwright.ps1 install
+
+      - name: Docker compose up
+        run: docker compose up -d
```
jenniferplusplus, line 52: If we use github actions built-in service containers, we won't have to manage a docker compose lifetime ourselves.
https://docs.github.com/en/actions/using-containerized-services/about-service-containers

jenniferplusplus (commented):
I'll pull this in a minute and see if I can see what's happening with razor rendering

Letterbook.Web.Tests/Letterbook.Web.Tests.csproj
ben-biddington: Ahm, good question, I'll check

ben-biddington:
> While using xUnit is also supported, we do not support running parallel tests

Mmm, seems like a poor limitation -- do you have more detail on that, @runewake2? 

I don't understand how [their thinks here](https://playwright.dev/dotnet/docs/test-runners#xunit-support) apply?

Does that imply that these do not work as expected?

```C#
public async Task InitializeAsync()
{
  _playwright = await Playwright.CreateAsync();
  Browser = await _playwright.Firefox.LaunchAsync(new BrowserTypeLaunchOptions { Headless = _opts.Headless });
}
```

I'd love to know more

Letterbook.Web.Tests/Letterbook.Web.Tests.csproj
ben-biddington: Okay looks like were okay without that reference

runewake2:
@ben-biddington The way xUnit works makes it great for unit tests but has some issues with long running e2e test cases run asynchronously.

* https://github.com/xunit/xunit/issues/2003
* https://github.com/xunit/xunit/issues/2111

I think this is why some other frameworks have moved to Nunit later.  https://github.com/hardkoded/puppeteer-sharp/pull/2276

I think you could get around this by disabling parallelism for these tests, but one impact from this is that we might see prolonged tests that might delay unit test results. Our current scope is small so it's hard to tell how much that will matter.

ben-biddington:
@runewake2 Ah well in practical terms then I'd just move them to another assembly so they don't block the pipe. I don't think there is an issue with xunit.

I assume there are parallel runners available to github actions? I mean so unit tests and larger tests can be run at the same time

runewake2:
There's an option for a matrix in GitHub Actions, that'd run things in parallel across different hosts. It might also be handy for coverage of distinct browsers (ex. a matrix over firefox, chrome, edge, etc)

The issue here is that xUnit seems to limit it's parallelism by *how many tests are running concurrently* not *how many tests that can be started concurrently*. What this means is that there is a scenario where every end to end test we have written is attempting to start a browser and execute concurrently because xUnit will schedule additional tests in the `await`s of async tests to optimize test execution. At scale I think this has implications for our test runners and could make it tough to run them.

ben-biddington:
@runewake2 Iiiiiiiinteresting! Okay let me move it to nunit

jenniferplusplus:
@ben-biddington If you'd like, we can pick this conversation up on discord. We have a small server for more real-time discussions between contributors. Let me know if you'd like an invite.

ben-biddington:
@jenniferplusplus okay sounds good, sorry for all the noise

.github/workflows/pull-request.yml
ben-biddington: Done using this one-liner, hope that's okay

hoverkraft-tech/compose-action@v2.0.1

jenniferplusplus:
There's nothing to apologize for! Chat is just easier for conversations. Check your mastodon DMs for the invite

jenniferplusplus (commented):
I had a look at the WebServerFixture, and I found a way to get it working. I feel a lot better about everything with that.

That said, I have a couple of issues.

1. This is a small thing, but could you exclude the style changes in the pull-request.yml action, and the contirbuting guide? I want to avoid thrashing changes in files like that. I'd be open to defining a style guide and tooling, but that should be its own concern.
2. Can you move the playwright tests to a dedicated CI job? I want to keep them out of the critical path until we have a strategy for this class of tests.

It also turns out it's kind of a hassle for me to run these tests. Installing the system dependencies is not straightforward on Fedora. I also don't love that this would make powershell a development dependency. The powershell scripts just wrap the Playwright dll, so it should be doable to replace them with c# scripts, in the future.

That's not a block, though. And I really appreciate you getting this over the line. If you can address 1 and 2, then I'm happy to get this merged 😄

ben-biddington:
Absolutely, leave it with me

.github/workflows/pull-request.yml
```diff
@@ -4,61 +4,79 @@ on:
...
+          dotnet user-secrets set "HostSecret" "$(openssl rand -base64 32)" --project Letterbook
+
+      - name: Start server
+        run: dotnet run --project Letterbook.Web &
+
+      - name: Install Playwright Browsers
+        run: pwsh Letterbook.Web.Tests/bin/Release/net8.0/playwright.ps1 install firefox
+
+      - name: E2E Test
+        run: BROWSER=firefox dotnet test Letterbook.Web.Tests.E2E --no-build
```
jenniferplusplus, line 67:
```suggestion
      - name: E2E Test
        run: dotnet test Letterbook.Web.Tests.E2E --no-build
        env: 
          BROWSER: firefox
          DOTNET_ENVIRONMENT: Development
```
I don't want to make any more changes while you're working, to avoid conflicts. This got the tests to run and pass for me locally. I had been doing it through the rider test runner, so I didn't notice the other env var

Letterbook.Web.Tests.E2E/Support/WebServerFixture.cs
```diff
@@ -118,10 +108,8 @@ private void CreateHost()
 			//
 			// We can work around some of the issues by overriding pages under Areas/IdentityPages/Account
 			.AddDefaultUI();
-		builder.Services.AddRazorPages(options =>
-		{
-			options.RootDirectory = razorPagesRootDirectory;
-		});
+		builder.Services.AddRazorPages()
+			.AddApplicationPart(Assembly.GetAssembly(typeof(Web.Program))!);
```
jenniferplusplus, line 112: oh, this is the actually relevant bit, btw

Letterbook.Web.Tests.E2E/Support/WebServerFixture.cs
```diff
@@ -135,7 +123,11 @@ private void CreateHost()
 		if (!app.Environment.IsDevelopment())
 		{
 			// Not sure if this works, with mixed Razor/WebApi
-			app.UseExceptionHandler("/Error");
+			// app.UseExceptionHandler("/Error");
+			app.UseExceptionHandler((applicationBuilder =>
+			{
+				Console.WriteLine("Error handler");
+			}));
```
jenniferplusplus, line 130: And this can be reverted. It's leftover debugging stuff (that didn't even work)

Letterbook.Web.Tests.E2E/Support/WebServerFixture.cs
```diff
@@ -20,7 +22,8 @@
 using Serilog.Enrichers.Span;
 using Serilog.Events;
 
-namespace Letterbook.Web.Tests.E2E.Support;
+// ReSharper disable once CheckNamespace
+namespace Letterbook.Web;
```
jenniferplusplus, line 26: Probably not actually necessary and could also be reverted

.github/workflows/pull-request.yml
ben-biddington: No that's fine you go for it and I'll pick it up in the morning :+1:

jenniferplusplus (approved):

PR #325, opened 2024-08-11
chore: sanitize posts

### LINKED ISSUE #303: html conversion from source

jenniferplusplus:
Adds content type-aware sanitization, and sanitizes posts along all create/update paths

Also, note the migration. When adding migrations, sometimes you need to help them out a little. Dotnet-ef will give a message like:
```shell
$ dotnet ef migrations add ContentTypeRequired
Build started...
Build succeeded.
An operation was scaffolded that may result in the loss of data. Please review the migration for accuracy.
Done. To undo this action, use 'ef migrations remove'
```

In this case, it was due to adding a NOT NULL constraint on the `ContentType` column. We can easily provide a default, using the `migrationBuilder`. This is something I want to get better about doing. In the past, I've just blown away migrations and started over, but the time is coming when that's not an option any more. So, consider this practice.

## Details
- html and plain text content types currently implemented
  - markdown would be an easy addition, on the backend
  - others are possible, MathML, perhaps

## Related
- closes #303

### DIFF
--- a/Letterbook.Adapter.Db/Migrations/20240811185620_ContentTypeRequired.cs
+++ b/Letterbook.Adapter.Db/Migrations/20240811185620_ContentTypeRequired.cs
@@ -1,0 +1,44 @@
+﻿using Microsoft.EntityFrameworkCore.Migrations;
+
+#nullable disable
+
+namespace Letterbook.Adapter.Db.Migrations
+{
+    /// <inheritdoc />
+    public partial class ContentTypeRequired : Migration
+    {
+        /// <inheritdoc />
+        protected override void Up(MigrationBuilder migrationBuilder)
+        {
+	        // Manually added to provide default values for existing null records
+	        migrationBuilder.UpdateData(
+		        column: "ContentType",
+		        table: "Content",
+		        value: "text/html",
+		        keyColumn: "ContentType",
+		        keyValue: null);
+
+            migrationBuilder.AlterColumn<string>(
+                name: "ContentType",
+                table: "Content",
+                type: "text",
+                nullable: false,
+                defaultValue: "",
+                oldClrType: typeof(string),
+                oldType: "text",
+                oldNullable: true);
+        }
+
+        /// <inheritdoc />
+        protected override void Down(MigrationBuilder migrationBuilder)
+        {
+            migrationBuilder.AlterColumn<string>(
+                name: "ContentType",
+                table: "Content",
+                type: "text",
+                nullable: true,
+                oldClrType: typeof(string),
+                oldType: "text");
+        }
+    }
+}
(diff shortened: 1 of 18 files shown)


Letterbook.Core/Extensions/DependencyInjection.cs
```diff
@@ -32,7 +34,25 @@ public static IServiceCollection AddLetterbookCore(this IServiceCollection servi
...
+					"a", "abbr", "address", "b", "bdi", "blockquote", "caption", "cite", "code", "col", "colgroup", "dd", "del", "dl", "dt",
+					"em", "i", "ins", "kbd", "li", "mark", "ol", "p", "pre", "q", "s", "strike", "strong", "sub", "sup", "table", "tbody",
+					"td", "tfoot", "th", "thead", "time", "tr", "tt", "ul", "var", "wbr"
+				},
+				AllowedAttributes = new HashSet<string>(){"href"},
+				AllowedCssClasses = new HashSet<string>(),
+				AllowedCssProperties = new HashSet<string>(),
+				AllowedAtRules = new HashSet<CssRuleType>(),
+				AllowCssCustomProperties = false,
+				AllowDataAttributes = false
```
jenniferplusplus, line 54: This is the HTML sanitization config. It applies everywhere we're sanitizing html input. Meaning, both from local posts (and profiles, etc, eventually) created by users, and remote posts received from peers. I'm very open to input on whether this is the right set of elements to allow. Note that the empty sets mean nothing is permitted in those categories.

Letterbook.Adapter.Db/Migrations/20240811185620_ContentTypeRequired.cs
```diff
@@ -0,0 +1,44 @@
...
+        /// <inheritdoc />
+        protected override void Up(MigrationBuilder migrationBuilder)
+        {
+	        // Manually added to provide default values for existing null records
+	        migrationBuilder.UpdateData(
+		        column: "ContentType",
+		        table: "Content",
+		        value: "text/html",
+		        keyColumn: "ContentType",
+		        keyValue: null);
```
jenniferplusplus, line 19: Filling out a default value in a column that used to be nullable

Letterbook.Core/PostService.cs
```diff
@@ -23,16 +23,18 @@ public class PostService : IAuthzPostService, IPostService
 	private readonly IPostAdapter _posts;
 	private readonly IPostEventPublisher _postEventPublisher;
 	private readonly IActivityPubClient _apClient;
+	private readonly IEnumerable<IContentSanitizer> _sanitizers;
 	private Uuid7 _profileId;
 	private IEnumerable<Claim> _claims = null!;
 
 	public PostService(ILogger<PostService> logger, IOptions<CoreOptions> options,
-		IPostAdapter posts, IPostEventPublisher postEventPublisher, IActivityPubClient apClient)
+		IPostAdapter posts, IPostEventPublisher postEventPublisher, IActivityPubClient apClient, IEnumerable<IContentSanitizer> sanitizers)
```
jenniferplusplus, line 31: Note the `IEnumerable<IContentSanitizer>`

If multiple factories for the same type are registered, this gives you all of them. So adding support for new content types just requires defining and registering new sanitizers

PR #330, opened 2024-09-25
fix: serializable claims

jenniferplusplus:
Use a serializable claim DTO in worker contracts

## Details
- It turns out `System.Security.Claims.Claim` doesn't serialize to json; who knew :woman_shrugging: :sweat_smile: 

## Related
- prep for #191

### DIFF
--- a/Letterbook.Workers/Consumers/TimelineConsumer.cs
+++ b/Letterbook.Workers/Consumers/TimelineConsumer.cs
@@ -34,30 +34,30 @@ namespace Letterbook.Workers.Consumers
 				case nameof(PostEventPublisher.Created):
 					return;
 				case nameof(PostEventPublisher.Deleted):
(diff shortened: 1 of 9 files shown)


Letterbook.Workers/Consumers/TimelineConsumer.cs
```diff
@@ -33,30 +34,30 @@ public async Task Consume(ConsumeContext<PostEvent> context)
 				case nameof(PostEventPublisher.Created):
 					return;
 				case nameof(PostEventPublisher.Deleted):
-					await _timelines.As(context.Message.Claims).HandleDelete(post);
+					await _timelines.As(context.Message.Claims.Select(c => (Claim)c)).HandleDelete(post);
```
warriordog, line 37: this Select statement gets repeated a lot. Could it be worth factoring out a helper method or something?

warriordog (approved):

PR #331, opened 2024-09-25
feat: make authority a queryable field on Profiles

jenniferplusplus:
The Authority field is the hierarchically ordered domain name from the FediId. That is, like `tld.domain.subdomain`. This is useful for doing fast queries to match a domain and all of its subdomains and ports, or exact matches on the full hostname.

## Details
- It's not critical because we don't have any production DBs, yet, but I would like to make the migration populate the correct data, which requires doing lots of string manipulation directly in SQL
- I could use some help with that, if anyone does that sort of thing

## Related
- prep for #191

### DIFF
--- a/Letterbook.Adapter.Db/Migrations/20240916044339_AddProfileAuthority.cs
+++ b/Letterbook.Adapter.Db/Migrations/20240916044339_AddProfileAuthority.cs
@@ -1,0 +1,59 @@
+﻿using Microsoft.EntityFrameworkCore.Migrations;
+
+#nullable disable
+
+namespace Letterbook.Adapter.Db.Migrations
+{
+    /// <inheritdoc />
+    public partial class AddProfileAuthority : Migration
+    {
+        /// <inheritdoc />
+        protected override void Up(MigrationBuilder migrationBuilder)
+        {
+            migrationBuilder.AddColumn<string>(
+                name: "Authority",
+                table: "Profiles",
+                type: "text",
+                nullable: false,
+                defaultValue: "");
+
+            migrationBuilder.CreateIndex(
+                name: "IX_Profiles_Authority",
+                table: "Profiles",
+                column: "Authority");
+
+            /**
+             * regex:
+             * \w+:\/\/				the scheme (ex. https://)
+             * (.*@)?				discarded capture group for the user info, just in case
+             * (\w+\.\w+[\.\w+]*)	capture group for the actual host name, which is what we want
+             * \/?.*				the port, path, query, and fragment portions
+             */
+            // \w+:\/\/(.*@)?(\w+\.\w+[\.\w+]*)\/?.*
+
+            // We have no servers, so no valuable data. So, we can probably skip this if necessary.
+            // But, this migration would actually leave existing profiles in a bad state.
+            // What we would need to do is extract, tokenize, reverse, and recombine the domain from the profile's
+            // fediId as part of the migration.
+            // Which is a _challenge_ to do in SQL.
+            // migrationBuilder.Sql(
+	            // """
+	            // UPDATE "Profiles" p
+	            // SET Authority=(SELECT "FediId" FROM "Profiles" p2 WHERE p2."Id"=p."Id")
+	            // """
+            // );
+        }
+
+        /// <inheritdoc />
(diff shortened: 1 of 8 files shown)


Letterbook.Adapter.Db/Migrations/20240916044339_AddProfileAuthority.cs
```diff
@@ -0,0 +1,59 @@
...
+            // But, this migration would actually leave existing profiles in a bad state.
+            // What we would need to do is extract, tokenize, reverse, and recombine the domain from the profile's
+            // fediId as part of the migration.
+            // Which is a _challenge_ to do in SQL.
+            // migrationBuilder.Sql(
+	            // """
+	            // UPDATE "Profiles" p
+	            // SET Authority=(SELECT "FediId" FROM "Profiles" p2 WHERE p2."Id"=p."Id")
+	            // """
+            // );
```
jenniferplusplus, line 44: Ideally the migration would include this, mostly as good practice for us. The basic requirement is that "Profiles".FediId is a URI. We need to extract the domain name, split it, reverse it, and recombine it to use as the value for "Profiles".Authority.

Ex: 'http://test.letterbook.example/actor/someid25string' -> 'example.letterbook.test'

The regex in the comments will extract the hostname (I think; not sure exactly what flavor of regex postgres uses), and I can easily write another regex to handle the tokenizing. If anyone feels adept at it, the part I could use help with is the array/string manipulation in SQL.

warriordog: personally, I'd be in favor of just not migrating this data since we don't have any real instances yet. I *can* help with the regex if we need, but I'd need to refresh my knowledge of Postgres' regex syntax.

warriordog (approved):

Letterbook.Adapter.Db/Migrations/20240916044339_AddProfileAuthority.cs
jenniferplusplus: I do want to build the habit of figuring out consistent migrations for when we do have real instances. But yes, I'm not going to let this hold up other development. It's not yet, so this is still fine.

jmelesky: The tricky thing here is reversing the array--I don't think postgres has a built-in for that. Are you open to having a SQL function be part of the migration?

IreneKnapp: `select (regexp_match('http://test.letterbook.example/actor/someid25string', '\w+://(.*@)?(\w+\.\w+[\.\w+]*)\/?.*'))[2];`

jenniferplusplus: @jmelesky I could be okay with that. 

@IreneKnapp thanks! The thing is I also need to reverse the tokens in the domain name. The idea is to support starts_with queries that capture ports and subdomains.

jmelesky: In that case (function courtesy of the postgres wiki: https://wiki.postgresql.org/wiki/Array_reverse):

`CREATE OR REPLACE FUNCTION array_reverse(anyarray) RETURNS anyarray AS $$
SELECT ARRAY(
    SELECT $1[i]
    FROM generate_subscripts($1,1) AS s(i)
    ORDER BY i DESC
);
$$ LANGUAGE 'sql' STRICT IMMUTABLE;

select array_to_string(array_reverse(string_to_array(regexp_replace(source_column, '\w+:\/\/(.*@)?(\w+\.\w+[\.\w+]*)\/?.*', '\2'), '.')), '.') from source_table;
`

jenniferplusplus: @jmelesky that works great on integration test data, so I'll take it, thank you!. Would you like a co-author on the commit?

jmelesky: Sorry, thought I'd already replied to this--I'm fine without attribution. Thanks!

PR #332, opened 2024-09-25
feat: outbound post federation

jenniferplusplus:
Federates posts out to peer AP inboxes, based on the mentions and specified audience.

I know it's big, sorry. It will be much easier to review and merge #329, #330, and #331 first. Unfortunately, this changed depends on all of them, so there's no good way to isolate just the federation related changes.

## Details
- This is probably less than minimum viable. It doesn't federate updates correctly, or deletes at all.
- It still needs interop testing, I haven't done that yet.
- But I added trace collection in integration tests, so now we can easily capture side effects

## Related
- major progress on #191


Letterbook.Core/PostService.cs
```diff
@@ -104,11 +104,20 @@ public async Task<Post> Draft(Post post, Uuid7? inReplyToId = default, bool publ
 			content.Sanitize(_sanitizers);
 		}
 
+		if (post.Audience.Contains(Audience.Public))
+		{
+			post.Audience = post.Audience.ReplaceWith(post.Creators.Select(Audience.Followers).ToHashSet());
```
warriordog, line 109: This seems like it replaces public posts with followers-only. Is that correct / intended?

warriordog (approved):

Letterbook.Workers.Tests/TimelineConsumerTests.cs
```diff
@@ -173,31 +173,6 @@ public async Task CanConsumeShared()
 		AuthzTimelineServiceMock.VerifyNoOtherCalls();
 	}
 
-	[Fact(DisplayName = "Should not add to timeline for malformed Shared events")]
```
jenniferplusplus, line 176: I removed this test because the changes to the message contract make this scenario essentially impossible

Letterbook.Core/PostService.cs
jenniferplusplus: Actually, no, it wasn't. That is a good catch. The idea is to add the followers audience if it wasn't specified, but we still want to retain the public audience.

Letterbook.IntegrationTests/PostsTests.cs
```diff
@@ -90,6 +126,26 @@ public async Task CanDraftNote()
 		Assert.NotNull(body.Id);
 		Assert.NotEqual(Uuid7.Empty, body.Id);
 		Assert.Equal(dto.Contents.First(), body.Contents.FirstOrDefault(), ContentComparer);
+
+		if(dto.Audience.Count > 0 || dto.AddressedTo.Count > 0)
+		{
+			const string expected = "urn:message:Letterbook.Workers.Contracts:ActivityMessage send";
+			Assert.Contains(expected, _host.Spans.Where(a => a.TraceId == traceId).Select(s => s.DisplayName));
```
jenniferplusplus, line 133: This is probably worth calling out explicitly. `_host.Spans` is the in-memory observable collection of spans produced by the application. This lets us use them in our tests, which makes it relatively easy to assert that things happened in the queue workers. This would have otherwise been nearly impossible.

Letterbook.IntegrationTests/Fixtures/HostFixture.cs
```diff
@@ -192,6 +221,14 @@ protected override void ConfigureWebHost(IWebHostBuilder builder)
 				{
 					options.ForwardAuthenticate = "Test";
 				});
+				services.AddOpenTelemetry()
+					.WithTracing(tracer =>
+					{
+						tracer.AddInMemoryExporter(_spans);
```
jenniferplusplus, line 227: Here, we configure tracing to export to an in-memory collection. This is actually in addition to the regular config, so if you have tempo running during your integration tests, you can also view those spans in grafana.

Letterbook.IntegrationTests/Fixtures/CollectionSubject.cs
```diff
@@ -0,0 +1,70 @@
...
+
+/// <summary>
+/// An Observable Subject that also implements ICollection, so it can be used as the backing collection for
+/// OpenTelemetry in-memory exporter
+/// </summary>
+/// <remarks>Using an Observable for this is useful because the Observable behavior makes it easy to act on spans as they arrive. A regular
+/// Collection doesn't do that, and we would need a lot more arbitrary waits and timeouts in order to be confident we had gotten the spans
+/// we need.</remarks>
+/// <typeparam name="T"></typeparam>
+public class CollectionSubject<T> : SubjectBase<T>, ICollection<T>
```
jenniferplusplus, line 15: This is an Observable subject that also implements ICollection. We use it as the backing store for the in-memory spans. This is necessary because the InMemoryExporter requires an ICollection, but normal collections won't let us react to new spans as they arrive. Since this is all happening asynchronously, that's an important feature.

Letterbook.Core.Tests/TheoryData.cs
```diff
@@ -0,0 +1,72 @@
+using System.Collections;
+
+namespace Letterbook.Core.Tests;
+
+/// <summary>
+/// A support class that allows us to have more idiomatic and type-safe ClassData classes to drive XUnit Theory tests
+/// ClassData implementations should inherit one of the generic derivatives of this class.
+/// </summary>
+public abstract class TheoryData : IEnumerable<object[]>
```
jenniferplusplus, line 9: TheoryData is a test utility that lets us have explicitly typed ClassData providers for XUnit theories.

jenniferplusplus (commented):
highlighting some new test capabilities

PR #337, opened 2024-10-18
feat: federate Activities for Posts

### LINKED ISSUE #191: Posts - Outbound federation

jenniferplusplus:
Federate AP Activities for common Post events.

## Details
- Fixed like/share/reply URIs on posts
- Added some additional trace based assertions on integration tests
- Started adding `.TagWithCallSite()` and `.TagWith(nameof(Function))` to queries for easier debugging
- Started to move some consumer filtering earlier in the pipeline, before the message body has been deserialized

## Related
- closes #191 

## Example Activity
```json
{
  "type": "Create",
  "@context": "https://www.w3.org/ns/activitystreams",
  "actor": "https://host.castle/actor/06e769cxim3gjfy0s1027ty87",
  "object": {
    "type": "Note",
    "@context": "https://www.w3.org/ns/activitystreams",
    "id": "https://host.castle/post/06e95an59v8758w08a7x5j62y",
    "attributedTo": "https://host.castle/actor/06e769cxim3gjfy0s1027ty87",
    "audience": "https://host.castle/actor/06e769cxim3gjfy0s1027ty87/followers",
    "context": "https://host.castle/thread/06e95an9sxb6shy4vv20qi1kx",
    "replies": {
      "type": "Collection",
      "@context": "https://www.w3.org/ns/activitystreams",
      "id": "https://host.castle/post/06e95an59v8758w08a7x5j62y/replies",
      "first": "https://host.castle/post/06e95an59v8758w08a7x5j62y/replies",
      "totalItems": 0
    },
    "content": "All this happened, more or less",
    "published": "2024-10-18T04:48:56.4225598",
    "likes": "https://host.castle/post/06e95an59v8758w08a7x5j62y/likes",
    "shares": "https://host.castle/post/06e95an59v8758w08a7x5j62y/shares"
  }
}
```


Letterbook.IntegrationTests/PostsTests.cs
```diff
@@ -129,48 +130,47 @@ public async Task CanDraftNote(Func<Profile, PostDto> generate)
...
-		// We expect the otherSpans timeout to trigger. This would mean no error spans were found, which is the desired result.
-		catch (TaskCanceledException)
-		{}
-		catch (OperationCanceledException)
-		{}
+		await Traces.AssertNoTraceErrors(traceId, _host.Spans);
+	}
+
+	[Fact(DisplayName = "Should draft a note without publishing")]
+	public async Task CanDraftNote()
```
jenniferplusplus, line 141: Note to self: this test and the one below seem to be confused :upside_down_face: 

Actual scenarios to cover:
1. draft without publishing (doesn't send any AP message)
2. publish an existing draft (does send AP message)

Letterbook.Workers/Consumers/OutboundPostConsumerDefinition.cs
```diff
@@ -0,0 +1,28 @@
...
+			await Task.FromResult(0);
+			return consumeContext.Headers.TryGetHeader(Headers.Event, out var header) &&
+			       header switch
+			       {
+				       nameof(PostEventPublisher.Published) => true,
+				       nameof(PostEventPublisher.Updated) => true,
+				       nameof(PostEventPublisher.Deleted) => true,
+				       nameof(PostEventPublisher.Shared) => true,
+				       nameof(PostEventPublisher.Liked) => true,
+				       _ => false
```
jenniferplusplus, line 24: Consumer Definitions allow us to customize the message pipeline for a consumer class. This specifically is filtering for the event types that OutboundPostConsumer is meant to handle. This happens before the message body is deserialized; at this point only the headers should have gone through deserialization. So it is both more efficient and somewhat simplifies the consumer implementation.

PR #349, opened 2024-12-02
feat: inbound federation

### LINKED ISSUE #190: Posts - Inbound federation
Handle Inbox activities related to Notes. This should cover Create, Update, and Delete activities. There may be others in the future, but this is enough to provide meaningful federation for now.

- [x] Add handling in the Inbox controller for Create, Update, and Delete activities
- [x] Map the objects carried by those activities to Posts[^1]
- [x] Pass off those Posts to the PostService for handling

[^1]: I don't think these mappers exist yet. Please reach out with any question at all.

jenniferplusplus:
Handles major inbound activities related to posts

## Details
- Added a crawl scheduler, to handle cases when inbound activities refer to unkown objects
  - should probably use scheduled crawls to handle all remote fetches, tbh
- The Receive* methods in `PostService` are pretty involved. I'd like them to be simpler, but I don't know if that's actually possible.
  - Need to keep an eye on them for performance, either way.

## Related
- Closes #190


jenniferplusplus (commented):
Also needs actual unit tests. Still, it's something.

PR #356, opened 2024-12-14
adr-11 design philosophy

jenniferplusplus:
ADR to discuss a design philosophy behind application features, and specific guidance on some upcoming UI features.


jenniferplusplus:
@mattly yeah, I think that could be very helpful. 

Tbh, I don't *really* know what I'm doing with this stuff. I'm trying to avoid a lot of mistakes that I've seen projects make. But I'm much more of an engineer than a designer, and I don't know what I don't know. Which is to say, feel free to volunteer things that you think are needed, because I probably don't know to ask for it.

PR #370, opened 2024-12-28
Fix building in Visual Studio by targeting multiple frameworks in Letterbook.Generators

### LINKED ISSUE #369: Build fails in Visual Studio
**Describe the bug**
Building the solution fails in visual studio. The build errors suggest that generated code is not available.

**Expected behavior**
Using the run command in Visual Studio would build and run the app.

**Additional context**
The generators project targets dotnet 8, just like the rest of the solution. This is a valid thing to do, and dotnet tooling is fine with it (dotnet CLI, JB Rider, VS Code, etc). But msbuild chokes on it. Generator projects should target netstandard2.0 to be compatible with msbuild. To my knowledge, there's no way to get visual studio to use dotnet tooling instead of msbuild.

- See also https://github.com/Letterbook/Letterbook/discussions/364

vivlim:
Visual Studio can't load code generators & analyzers that target net8.0 (https://github.com/dotnet/roslyn/issues/47087, https://github.com/dotnet/roslyn-analyzers/issues/7302).
With this change, `Letterbook.Generators` will now be built twice, targeting both `netstandard2.0` and `net8.0`, and the `netstandard2.0` build will be referenced when building inside of Visual Studio.

To validate this change, I built the solution with Visual Studio and dotnet on a x64 windows machine. I ran all the tests in VS and they passed, excluding `Letterbook.IntegrationTests` and `Letterbook.Web.Tests.E2E` which failed and I assume will need additional setup.

## Details
- `Letterbook.Core`'s reference to `Letterbook.Generators` is unfortunately significantly more complex now.
  - When building inside VS, it's referenced *twice*; the `netstandard2.0` version is used for source generation, and the `net8.0` version is the assembly that's actually referenced by `Letterbook.Core`.
  - When building outside of VS, it only references the `net8.0` version, which is the same behavior as before.
- When building `Letterbook.Generators` targeting `netstandard2.0`, `ITypedId<T>.FromString` is not built since there's no runtime support for `static abstract` members in interfaces; fortunately it's not needed during source generation time.
- I tried a couple alternatives that had a larger footprint and felt messier:
  - Moving `ITypedId<T>` into `Letterbook.Core`, this would either require changing namespaces or introducing an inconsistency by having a type in the `Letterbook.Generators` namespace living in the `Letterbook.Core` project.
  - Specifically including `ITypedId.cs` from `Letterbook.Generators` in `Letterbook.Core.csproj`. Since this depended on `.cs` file paths it felt unacceptably brittle.

## Related
- Fixes #369

### DIFF
--- a/Letterbook.Core/Letterbook.Core.csproj
+++ b/Letterbook.Core/Letterbook.Core.csproj
@@ -35,8 +35,14 @@
       <PackageReference Include="OpenTelemetry.Instrumentation.Http" />
     </ItemGroup>
 
-    <ItemGroup>
-      <ProjectReference Include="..\Letterbook.Generators\Letterbook.Generators\Letterbook.Generators.csproj" OutputItemType="Analyzer" ReferenceOutputAssembly="true"/>
+    <ItemGroup Condition="'$(MSBuildRuntimeType)' != 'Full'"> <!-- Building using dotnet tooling. -->
+      <ProjectReference Include="..\Letterbook.Generators\Letterbook.Generators\Letterbook.Generators.csproj" OutputItemType="Analyzer" ReferenceOutputAssembly="true" SetTargetFramework="TargetFramework=$(TargetFramework)" />
+    </ItemGroup>
+    <ItemGroup Condition="'$(MSBuildRuntimeType)' == 'Full'"> <!-- Building inside Visual Studio. -->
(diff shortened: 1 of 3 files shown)


jenniferplusplus:
Thanks for doing this!

PR #374, opened 2024-12-31
feat: safety & moderation data models

### LINKED ISSUE #371: Moderation related data models
- [ ] Reports
- [ ] Moderation Policies
- [ ] Peers (aka, other servers)

jenniferplusplus:
Adds data models to support putting restrictions and federation limits on peers and profiles.

I think one thing it's useful to call out is that I thought that moderation should be performed by accounts, rather than profiles. Using profiles would have opened up the possibility that remote users could perform moderation under their established name. Which is an interesting idea, and might make it easier to bring in outside help when it's needed. But I decided it's more important that moderation be locally situated, and connected to people, rather than "identities" or personas. Accounts are the closest thing to that concept for us. So, guest moderators can just make an account on the server, exactly as I'm sure they do now in all other fediverse software.

## Related
- closes #371
- See also #366

## Models
image

For clarity, note that Accounts, Profiles, Posts, and Threads already exist. The new models have various relationships to these concepts.


jenniferplusplus:
Thinking about this some more, I realize that there's no clear representation of an appeal. Or really even of an action or decision, for that matter. Any thoughts on how much those things should or could be included in the data model?

jenniferplusplus:
Another question I don't have an obvious answer for: what's the best way to handle receiving multiple reports for the same event?

My original thinking was to have the concept of a case, and that reports would be added to cases and cases are what would be investigated, discussed, and resolved. In that model, those duplicate (corroborating?) reports could all be attached to the same case. I haven't added that concept to the domain model because after talking with IRL moderators, I'm not sure it's the right thing anymore. Cases, as a container for multiple reports and with a singular conclusion, seem better suited to coordinating across large groups of people. But the actual challenge is more about coordinating across long(ish) spans of time.

And I'm not sure what the organizing concept is for long running, possibly open ended coordination. The kind of problem this would need to solve is to represent when this person has a stalker, or that person has spurred harassment campaigns before. Things that are not so much events as they are patterns or situations. Things that are hard to really grasp without a historical view.

But that still leaves me without a clear answer for how to facilitate responding to multiple reports.

jenniferplusplus:
Here's my current line of thinking.

## Handling duplicates
Allow moderators to merge reports together. Essentially, transfer all of the notes/comments/etc from one report to another, and then leave the empty report as a pointer to the combined one.

## Handling long running concerns
1. Create situation report, or something along those lines. This would support collecting reports, posts, notes, etc, along with a human-written summary of the situation. These would exist for the purpose of record-keeping; not as something that needs a response and conclusion.
2. Labels. Just user-defined text annotations, which would support searching, grouping, aggregates, and filters. These could be attached to more-or-less anything.

Letterbook.Core/Models/Restrictions.cs
```diff
@@ -0,0 +1,21 @@
...
+	DenyPrivate,
+	/// Reject shared posts
+	DenyShare,
+	/// Reject posts in public/shared audiences
+	NoPublic,
+	/// Hide posts from non-followers
+	NoDiscovery,
+	/// Discard all posts (permit profile moves and updates)
+	Ignore,
+	/// Discard all messages
```
jenniferplusplus, line 19:
```suggestion
	/// Hide posts from non-followers
	NoDiscovery,
	/// Discard all images, audio, video, quote posts, etc
	NoAttachments,
	/// Discard all links
	NoLinks,
	/// Discard all posts (permit profile moves and updates)
	NoPosts,
	/// Discard all messages
```

Letterbook.Core/Models/Restrictions.cs
```diff
@@ -0,0 +1,21 @@
...
+	Defederate
```
jenniferplusplus, line 20:
```suggestion
	/// No restrictions
	None,
	/// Click-through warnings on posts/profiles, require approval for follow requests
	Warn,
	/// Hide posts from non-followers
	LimitDiscovery,
	/// Reject private messages
	DenyPrivate,
	/// Reject posts in public/shared audiences
	DenyPublic,
	/// Reject shared posts
	DenyShare,
	/// Discard all images, audio, video, quote posts, etc
	DenyAttachments,
	/// Discard all links
	DenyLinks,
	/// Discard all replies (permit top-level posts and self-replies)
	DenyReplies,
	/// Discard all posts (permit profile moves and updates)
	DenyPosts,
	/// Discard all messages
	Defederate
```

jenniferplusplus:
@mattly As I have it organized right now, a report can have 0 or more subjects (profiles that are a concern of the report) and 0 or more posts. This is maybe a little bit easier to see in code than in the db schema, because those relations are many-many relations with an intermediate join table. For convenience, here's the code:
```csharp
public class ModerationReport
{
	public ModerationReportId Id { get; set; }
	public Uri? FediId => Context.FediId;
	public required ThreadContext Context { get; set; }
	public ICollection<Account> Moderators { get; set; } = new HashSet<Account>();
	public ICollection<ModerationPolicy> Policies { get; set; } = new HashSet<ModerationPolicy>();
	public Profile? Reporter { get; set; }
	public ICollection<Profile> Subjects { get; set; } = new HashSet<Profile>();
	public ICollection<Post> RelatedPosts { get; set; } = new HashSet<Post>();
	public DateTimeOffset Created { get; set; } = DateTimeOffset.UtcNow;
	public DateTimeOffset Updated { get; set; } = DateTimeOffset.UtcNow;
	public DateTimeOffset Closed { get; set; } = DateTimeOffset.UtcNow;
	public IEnumerable<ModerationRemark> Remarks { get; set; } = new List<ModerationRemark>();
}
```

Which is to say, a flag activity could reference any number of actors, notes, or most other kinds of objects[^1], in any combination, and we would be able to have a coherent representation of that in our domain models. The report is about any and all of those things, collectively. So, if the mods merge two reports together, I'm imagining the final result would simply include the set union of all of the related objects from both reports.

That also creates the possibility that a user could create a more robust report in the first place. For instance, a single report that several other accounts are doing tag-team harassment could just directly reference those several accounts and some illustrative posts.

> What would you call the merged entity?

I think it would just be another `ModerationReport`. I would add some relations to track the merge destinations. Like:
```csharp
public ModerationReport? MergedInto {get; set;}
public ICollection<ModerationReport> MergedFrom {get; set;}
```

A report that is the combination of several other reports would reference them as `MergedFrom`

It will be tricky to make this legible to mastodon without losing a lot of fidelity. But that's a problem for future (like 2 - 4 weeks) me.

[^1]: AP has basically no restrictions on anything. It would be a valid AP document for a Flag to reference collections, other activities, links, or even itself. Most of which would be absurd, and none of which would be understood by any backend at all. But AP only concerns itself with semantic expressiveness, not logical comprehensibility.

jenniferplusplus:
@mattly 
> the idea that a profile/account with a certain label attached should force a set of additional behavior

I'm not sure how to actually do that. I guess in part because there's no concept of actions or resolutions in this model, yet. But also in part because it seems very sensitive to the particulars of a given workflow.

My first thought is to just let that be communicated with labels, and otherwise leave it to humans to decide what to do.

But, do you have any thoughts on what the nouns are that make up a decision/action/resolution model?

Some scenarios, off the top of my head:
1. decide the report is not valid (like the caturday-hater)
2. decide the report is malicious (like report bombing)
3. decide the report is valid, but doesn't warrant action
4. decide the report is valid and act on it

And then the kind of actions you might take:
1. issue a warning
2. apply or remove a tag/label (like private notes)
3. delete or modify a post
5. (in the future) add or modify a label on something (like bsky labellers)
6. apply temporary or permanent restrictions
7. apply temporary or permanent protections?
8. apply temporary or permanent federation limits
9. modify policies

Which is pretty wide ranging

jenniferplusplus:
> I also think it would be useful to have some notion of "resolved" through a boolean flag or maybe a state enum, simply for listing purposes.

The report has a `Closed` date, which is a similar pattern as with posts and profiles, where they have a deleted date. The simple case is that we would query for `WHERE closed = null` to get open reports. (For posts and profiles, I wanted to retain the record so that we can display tombstones rather than fragmenting threads. It also preserves context for investigating reports.)

> at the same time, maybe it's not a resolution necessarily, because multiple actions could be taken across time.

Yeah. While I was writing those lists, what came to mind is that I'm almost imagining an audit log (a feature I definitely want). And this seems like more of the same. The thing is, it also suggests an association from audit log entries to moderation reports, and that's hard to do without making the work flow really cumbersome.

PR #375, opened 2025-01-02
chore: web scaffold

### LINKED ISSUE #373: Scaffold initial web views
Set up initial razor layouts, css, build, tooling, etc

jenniferplusplus:
Scaffold up something closer to what the web views should actually look like

## Details
- Removed bootstrap
- Using PicoCSS, and currently zero javascript
- Takes colors and fonts that are already established for the docs site
- Mostly just sets up the default layout; pages are essentially unaffected
  - Uses CSS grid for the main layout
  - Not responsive, yet :sweat_smile:
- Set up library manager for web assets

## Related
- closes #373

## Preview
image

with css grids highlighted
image


jenniferplusplus:
I'll need to add back the login/sign links. Otherwise, there would be no good way to get to those pages and actually log in, which is useful.

PR #401, opened 2025-02-17
feat: moderator api

### LINKED ISSUE #395: Add API controllers for reports/moderators

jenniferplusplus:
Adds moderation core service, and reports API. Reports support the basic moderation workflows of opening reports about undesirable profiles/posts, managing moderation policies, and reviewing reported content.

## Details
- Adds core service
- Adds API controller
- Includes fixes to OpenAPI docs from #400
- And the authz updates from #402 

## Related
- closes #395


jenniferplusplus:
Summarizing some discussion that happened in chat:

- Renamed some of the new db tables to clarify the relation, instead of using the direct model names. New names are `ReportSubjectPost` and `ReportSubjectProfile`
- It's conceivable that we would want to enable reporting other kinds of objects in the future. For example, lists/starter packs.
- We don't make a strong distinction between locally created reports, and reports received from peers. However, we're likely to have more information with local reports.
- It is possible to report multiple posts, and also multiple profiles together. The idea is to enable reporting complex situations involving multiple people. Sock puppets, or group harassment are scenarios I specifically had in mind.
- The ThreadContext that's included on reports is intended to capture communication with the author, subjects, and/or remote peer moderators. We don't have a capability to actually perform that communication, yet. But that's the idea.

Issue #352, opened 2024-12-10
Home feed view UI


jenniferplusplus:
The basic purpose of the home view is to let members scan recent posts, and keep tabs on what's new with the people they follow. The home view uses the standard layout, and has two main sections:

- [ ] #416
- [x] a feed

## Conversations
Conversations are one of the major distinguishing features in Letterbook. The idea is to make recent active threads involving people you follow visible and obvious. My thinking is that these threads are likely to be some of the most interesting threads for a member. They should be clearly distinct from the feed, and they are not ordered chronologically.

## Feeds
Feeds are the typical scrollable list of posts that most social media centers on. It's generally reverse chronological order. The default home feed is the full set of posts to any audience you're a member of. Typically, that means all of the public/followers posts from everyone you follow, and every post they share. We would likely also implement lists, or home feed customization to exclude shares or replies. It's also conceivably possible we would eventually implement ATProto federation, and ATP-style feed generators. In any case, the UI concern is to render a list of posts, including the author(s), and create/edit date.

jenniferplusplus:
Rough sketch of the home view

![Image](https://github.com/user-attachments/assets/f86d1498-3b30-478d-885e-1420f5576b37)

(made in https://excalaidraw.com. Scene data is embedded.)

PR #414, opened 2025-03-30
feat: home feed view

jenniferplusplus:
Add an initial home page for the app

>[!Tip]
> Run the Letterbook.Web.Mocks project to get some quick mock data to fill out the UI

## Details
- default collapsing 2-column layout
- authenticated users get their default timeline
- unauthenticated users should get some kind of server info, tbd

## Related
- #352

## Views
Desktop

Mobile


jenniferplusplus:
Still a work in progress. Right now, I'm trying to get the sidebar header and footer to both be sticky, and also layout the html in the expected way. Like this:

```html
<body>
  <header>...</header>
  <body>...</body>
  <footer>...</footer>
</body>
```

However, they don't stick like I want at the moment. Here's a slightly scrolled example. And if you keep scrolling, the footer will collide with the header and both are pushed off screen.
Screen Shot 2025-03-30 at 16 55 12

jenniferplusplus:
also, please forgive the emojis, I don't have good icons yet

Source/Letterbook.Web/wwwroot/css/app.css
```diff
@@ -1,145 +1,369 @@
...
+    grid-template-rows: auto;
+    height: auto;
 
-    footer {
-        grid-area: footer;
+    & > header footer,
+    & > footer {
+        grid-area: header;
         font-size: 0.75rem;
+        position: fixed;
```
jenniferplusplus, line 13: need to figure out why making this sticky isn't working, because making if fixed also causes a bunch of problems

Issue #353, opened 2024-12-10
Profile view UI


jenniferplusplus:
Profile page should use the standard layout (`Layout_.cshtml`, it's the default if you don't specify a different one).

At a high level, there should be two sections to the page:

- [x] biographical info
- [x] posts

Biographical info allows people to control some elements of how they present themselves in the fediverse. That includes the display name, profile picture, summary, custom info, etc. It also provides a measure of confidence that this profile represents the person it appears to be. The biggest one is rel=me verification of links in custom info. But, that also includes the followers/following count, and age of the profile.

Profile pages also allow the user to manage their relationship and interactions with the subject of the page. That means you can do things like follow or unfollow from here, as well as block, unblock, mute, unmute, etc.

Posts should be displayed using a common reusable widget, as there are at least 3 other pages that will display posts across the app (feeds, threads, reports). They should all have consistent presentation. The specific design of that presentation is not done, as of now.

jenniferplusplus:
Take this as a general guide for the profile page layout.

![Image](https://github.com/user-attachments/assets/6fa2469d-9fc0-431b-bda5-245d2dfb98a1)

PR #418, opened 2025-04-15
feat: projected and estimated values

jenniferplusplus:
Configures projections to begin to support querying complex and aggregate values. Also adds simple estimates, to track values that may be reported by peer instances.

## Details
- Adds projectable library, and some initial projectable methods
  - Projections aren't all that useful on basic model objects, but they can be used to populate values on mapped or projected objects
  - Automapper configs can sometimes be used as projection targets, which is helpful, and can make code reuse easier.
  - But only if the mapping doesn't use several techniques that we do tend to rely on. As a rule of thumb, prefer `ConvertUsing(expr)` methods, rather than value mappers or resolvers.
  - See the AutoMapper docs for details on restrictions when using projectionshttps://docs.automapper.org/en/stable/Queryable-Extensions.html#supported-mapping-options
- Estimated fields may or may not be populated
  - On remote objects, check if they are and use the reported value if we have one
  - We don't currently fetch those values from anywhere.
  - They likely need to be crawled from links on peer profiles.
  - Post-related estimates may sometimes be available on the object, without crawling the relevant collections.
  - But we would still likely crawl many of those collections, in many scenarios.

## Related
- prep for #353

### DIFF
--- a/Source/Letterbook.Adapter.Db/EntityConfigs/ConfigureProfile.cs
+++ b/Source/Letterbook.Adapter.Db/EntityConfigs/ConfigureProfile.cs
@@ -21,5 +21,8 @@ public class ConfigureProfile : IEntityTypeConfiguration<Models.Profile>
 			.WithOne(relation => relation.Follows)
 			.IsRequired();
 		builder.Property(profile => profile.Restrictions).HasColumnType("jsonb");
+		builder.Property(profile => profile.FollowersEstimate).HasDefaultValue(-1);
+		builder.Property(profile => profile.FollowingEstimate).HasDefaultValue(-1);
+		builder.Property(profile => profile.PostsEstimate).HasDefaultValue(-1);
 	}
 }
--- a/Source/Letterbook.Core/Models/Mappers/PostMappings.cs
+++ b/Source/Letterbook.Core/Models/Mappers/PostMappings.cs
@@ -68,17 +68,3 @@ public class PostMappings : AutoMapper.Profile
 
 	private static Post NewPost(CoreOptions options) => new(options);
 }
-
-public class PublicKeyPemConverter : IValueResolver<SigningKey, PublicKeyDto, string>
-{
-	public string Resolve(SigningKey source, PublicKeyDto destination, string destMember, ResolutionContext context)
-	{
-		return source.Family switch
-		{
-			SigningKey.KeyFamily.Rsa => source.GetRsa().ExportSubjectPublicKeyInfoPem(),
-			SigningKey.KeyFamily.Dsa => source.GetDsa().ExportSubjectPublicKeyInfoPem(),
-			SigningKey.KeyFamily.EcDsa => source.GetEcDsa().ExportSubjectPublicKeyInfoPem(),
-			_ => null!,
-		};
-	}
-}
--- a/Source/Letterbook.Core/Models/Mappers/ProfileMappings.cs
+++ b/Source/Letterbook.Core/Models/Mappers/ProfileMappings.cs
@@ -1,5 +1,8 @@
-﻿using AutoMapper;
+﻿using System.Text;
+using AutoMapper;
 using Letterbook.Core.Models.Dto;
+using Org.BouncyCastle.Utilities.IO.Pem;
+using PemWriter = Org.BouncyCastle.OpenSsl.PemWriter;
 
 
 namespace Letterbook.Core.Models.Mappers;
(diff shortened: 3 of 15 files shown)


Tests/Letterbook.IntegrationTests/DataAdapterTests.cs
```diff
@@ -70,7 +81,20 @@ public async Task QueryFromProfileAndInclude()
...
 		Assert.NotEqual(Uuid7.Empty, profile.GetId());
 	}
+
+	[Fact(DisplayName = "Query profile projection fields")]
+	public async Task QueryProfileProjections()
+	{
+		var id = _profiles[7].Id;
+		var adHocProjection = await _adapter.Profiles(id).Select(e => new {Followers = e.FollowersCount}).FirstAsync();
+		var automapProjection = await _adapter.Profiles(id).ProjectTo<FullProfileDto>(_mappingProvider.Profiles).FirstAsync();
+		var noProjection = await _adapter.Profiles(id).FirstAsync();
```
jenniferplusplus, line 94: This is probably the best example of how those projections can be useful

Source/Letterbook.Core/Models/Mappers/PostMappings.cs
```diff
@@ -67,18 +67,4 @@ public PostMappings(IOptions<CoreOptions> options)
 	}
 
 	private static Post NewPost(CoreOptions options) => new(options);
-}
```
jenniferplusplus, line 70: This logic moved to ProfileMappings. It never made sense here

Source/Letterbook.Core/Models/Mappers/ProfileMappings.cs
```diff
@@ -9,16 +12,40 @@ public class ProfileMappings : AutoMapper.Profile
...
+		CreateMap<SigningKey, PublicKeyDto>()
+			.ConvertUsing(src => new PublicKeyDto
+			{
+				Label = src.Label,
+				Family = src.Family.ToString(),
+				PublicKeyPem = PemStringBuilder(src.PublicKey).ToString().Trim(),
+				Created = src.Created,
+				Expires = src.Expires,
+				FediId = src.FediId
+			});
```
jenniferplusplus, line 39: Replaced the value converter with a convert expression, to support projections

Source/Letterbook.Core/Models/Mappers/ProfileMappings.cs
```diff
@@ -9,16 +12,40 @@ public class ProfileMappings : AutoMapper.Profile
...
+	}
+
+	private static StringBuilder PemStringBuilder(ReadOnlyMemory<byte> publicKey)
+	{
+		var builder = new StringBuilder();
+		using TextWriter writer = new StringWriter(builder);
+		var pemWriter = new PemWriter(writer);
+		var pem = new PemObject("PUBLIC KEY", publicKey.ToArray());
+		pemWriter.WriteObject(pem);
+		return builder;
```
jenniferplusplus, line 49: TBH, this is a nicer implementation, anyway. And, this method reduces our reliance on the System.Security.Cryptography implementations. If this can replace our uses of `AssymetricAlgorithm.ExportSubjectPublicKeyInfoPem()`, then we can also switch to BouncyCastle for key generation, and we'd be able to use ed25519 signatures.

Source/Letterbook.Adapter.Db/EntityConfigs/ConfigureProfile.cs
```diff
@@ -21,5 +21,8 @@ public void Configure(EntityTypeBuilder<Models.Profile> builder)
 			.WithOne(relation => relation.Follows)
 			.IsRequired();
 		builder.Property(profile => profile.Restrictions).HasColumnType("jsonb");
+		builder.Property(profile => profile.FollowersEstimate).HasDefaultValue(-1);
+		builder.Property(profile => profile.FollowingEstimate).HasDefaultValue(-1);
+		builder.Property(profile => profile.PostsEstimate).HasDefaultValue(-1);
```
jenniferplusplus, line 26: Setting default values in the entityConfig _should_ allow EFCore to tell the difference between previously set values and newly scaffolded ones, so the set values are preserved.

jenniferplusplus (commented):
Some of the more interesting bits

Source/Letterbook.Core/Models/Mappers/ProfileMappings.cs
jenniferplusplus: Note, this is tested here https://github.com/Letterbook/Letterbook/blob/1c9250d183faf30f30a1954d3bc6f74758888dde/Tests/Letterbook.Core.Tests/MapperTests.cs#L291

The produced PEM string is byte-for-byte equal to the SSC methods.

Issue #354, opened 2024-12-10
Thread view UI


jenniferplusplus:
The thread view serves to allow a member to follow and participate in a conversation. It's important to note that neither ActivityPub nor ATProto has any concept of a conversation. Both only conceive of post-like records which may or may not be in reply to another post-like record. For better or worse, this results in the formation of unbounded reply trees. Displaying reply trees is hard. There's not really a good solution, but I hope this strikes a balance among our various concerns:

1. Show posts in context. This is the primary goal of a thread view. That context includes, at least, the reply chain up to the root of a thread, and the relevant post's siblings and children.
2. Make existing replies readily visible. We want to give people the best chance to see whether someone has already made the point they want to make. My hope is that will lead toward interactions that are more like a conversation, and less like a chanting crowd.
3. Don't be overwhelming. Threaded reply trees can be very large, complex, and wide-ranging. Navigating that takes real mental effort. We should present a relevant subset, and make it easy to move to different subsets, rather than try to flatten the whole thing into a single screen.

jenniferplusplus:
This is the part of the UI design that I think is most unsettled right now. But, we have to start somewhere, so here's a mock up.

![Image](https://github.com/user-attachments/assets/5eac8cd7-29fe-47a6-a704-efc419f52c6a)

jenniferplusplus:
I'm going to call this done enough. Future enhancements should happen under a new issue

PR #422, opened 2025-05-27
feat: threads

jenniferplusplus:
Initial, partially functional, thread page view. When opening a post, this page loads the post in context of its thread, and allows you to reply. It's all still pretty minimalistic and naive, but basically functional.

## Details
- thread page
  - single thread
  - plus sibling and children of the selected post
- reply form
  - works, needs error handling tho
- does not include:
  - reply addressing or visibility
  - a good way to open this page (nothing links here yet)
  - reply/like/share counts
  - paging

## Related
- re #354

Screenshot of the reply form

### DIFF
--- a/Source/Letterbook.Web/Pages/Thread.cshtml
+++ b/Source/Letterbook.Web/Pages/Thread.cshtml
@@ -17,7 +17,8 @@
             </footer>
         </article>
     }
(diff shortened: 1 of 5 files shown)


jenniferplusplus:
big pic here, but this is what a non-trivial thread looks like
image

Issue #215, opened 2024-04-22
Add Account Management to UI


mattly:
## What is the goal of this feature?
(derived from #141 ) - Have affordances in the UI to manage various aspects of the user's _account_, which is distinct from their _profile_.

## Describe the requirements & constraints
For our purposes I'm going to stay focused on what's under the "Account Settings" portion of the Mastodon UI, which has the following features: 

- Update email address. It seems doing this requires entering your password for security
- Update password. It seems doing this requires entering your current password for security.
- Display of currently logged-in sessions, with IP address, device type, and last activity, and the ability to revoke each session
- Move to a different account
- Move from a different account
- Delete account
- Manage 2FA via Authenticator App
- Manage 2FA via Security Keys
- List of authorized apps, with the ability to revoke each. (mine is .... 15 pages long, yikes)

What of this is necessary / nice-to-have / unnecessary for Single-User preview? If it's not necessary, is it something we want to support eventually?

My hunch is that for SUP we can ignore Move to / from different account and delete account. 

What about 2FA?

Management of sessions & authorized apps is important as well. Should this be included in SUP? can we put that off?

jenniferplusplus:
@mattly Coming back to this, I think the necessary account related things for SUP are just:
- _invites_
- signup
- login
- log out
- update email
- and update password

Note the addition of invites, because that's how I planned to enable closing registrations. You would make invites mandatory and then just not ever create any invites for anyone else. It's also the solution to the chicken-egg problem around creating the first admin account. The system will generate an invite code on first launch and print it in the logs/console, so the admin can securely create an account using that code.

I think we can call 2FA and session management nice to have for SUP. Definitely not necessary.

Moving "accounts" is actually going to mostly be a profile level concern in Letterbook (and also a post-SUP concern). The interaction with accounts would come in when creating a new profile to receive the inbound migration. Deleting accounts will have both Account and Profile components. We can work through that more in the future. Also not necessary for SUP. In fact I'd rather hold it back explicitly, because I wouldn't want to inadvertently encourage people to move their real fediverse identities onto a service that is likely to still be very unstable.

jenniferplusplus:
The invite backend stuff doesn't exist yet, but everything else does for what's on the necessary list

cederbs:
Alright, coming in to summarize what I've read from this, but spinning more of the discussion towards the experience rather than implementation:

### Goal
When someone creates an account, they expect there to be tools in place to keep their account secure and accessible. 

[...]

PR #431, opened 2025-09-20
feat: account page

jenniferplusplus:
Adds basic account management, allowing a member to change their email and password

## Details
- Includes a token-based email confirmation feature, embedded in a link
  - We don't have email support yet, so for now we put the link in the page

## Screenshots
image

image

image

## Related
- re: #215


cederbs:
@jenniferplusplus 
Some minor copy changes. I'd recommend using sentence case over title case for the headers, and avoiding "your" in labels just because that's implied. 

I also changed the page headline to 'Settings' since that's how it's referred to in the righthand navigation.

My only hot take is that—if possible—I think doing away with 'confirm password' and instead opting for a visibility toggle to be able to review the password would probably be more favorable here, but your implementation is incredibly common and shouldn't cause any issues.

Screenshot 2025-09-21 at 10 52 46 AM

jenniferplusplus:
Thanks for the feedback!

I'm trying to push the UI as far as I can without any javascript, and I don't think there's a way to toggle password visibility without that. We'll eventually add it as a progressive enhancement, but I want to get a solid functional base established first. But the rest is doable

Issue #385, opened 2025-01-16
Webfinger search


jenniferplusplus:

jenniferplusplus:
Actually, this just duplicates #437

PR #470, opened 2026-01-27
Issue/385 implement web finger search

ben-biddington:
Made a start on [implementing webfinger](https://github.com/Letterbook/Letterbook/issues/385) at something like

```
/lb/v1/.well-known/webfinger?resource=acct:{profile.Handle}@{profile.Authority}
```

* Request invokes `ProfileService.FindProfiles` after parsing the handle out of the `resource`.
* Based on [webfinger](https://datatracker.ietf.org/doc/html/rfc7033#section-4.2) and [The 'acct' URI Scheme](https://datatracker.ietf.org/doc/html/rfc7565).

Still a few questions in the tests there.


ben-biddington:
Okay the latest is that some of the tests pass against `/.well-known/webfinger` with [one small change that allows test profiles to be found](5c6efea27af3cd374fbc9c98a3606b9711298289).

That change may be useful for local development at least because otherwise test profiles are never returned.

Returning HTTP 400 from `WebFingerProvider` may not be possible and the failures all represent bad requests.

Issue #437, opened 2025-10-12
Search API


jenniferplusplus:
# Purpose

Lookup profiles by handle. Check the local data set and fallback to webfinger if we don't have it.

## Steps

- [x] Implement `IProfileSearchProvider` and `IGlobalSearchProvider` in the data adapter (should parse queries for `@` handles)
- [x] Call appropriate search providers from a search service
- [ ] Call the search service from a search API controller

## Related

- Other issues

jenniferplusplus:
Yes, I was imagining a search API. If you have a look in `Letterbook.Core.Adapters`, I defined a few search provider interfaces. It probably makes sense to align the api with those interfaces. So, something like this

```
GET /lb/v1/search?q={query}
GET /lb/v1/search_profiles?q={query}
GET /lb/v1/search_posts?q={query}
... maybe other categories in the future
```

Although I'm open to other ways to organize it. Maybe a query param, like `q={query}&kind=post|profile`, where `kind` is optional. Whatever seems best.

The service backing that should depend on an IEnumerable of the providers. That way, dependeny injection will provide all of the registered implementations. And then the service can call them each and concatenate the results. That would look roughly like this

```scharp
// other kinds tbd
public class SearchService(IEnumerable<IProfileSearchProvider> profileSearchProviders)
{
  public SearchService(IEnumerable<IProfileSearchProvider> profileSearchProviders) // other 

  // or IAsyncEnumerable, if we can get that to work
  public Task<IEnumberable<Profile>> SearchProfiles(string query, cancelationToken cancel) {
    foreach (provider in profileSearchProviders)
    {
      var results = await provider.searchProfiles(query, cancel)
      // do stuff with results
    }
  }
}
```

The database adapter should also implement those same search providers. That can be done later, or as part of this issue.

jenniferplusplus:
@ben-biddington I'm going to be working on other parts of search today and tomorrow. If you haven't built a core search service yet, I can take that. That would mainly leave the API controller, and implementing the search interfaces in the db adapter, if you're up for that

ben-biddington:
Sounds like a plan, although now it's me on call for the weekend 🫠

jenniferplusplus:
[...]

PR #482, opened 2026-02-17
Issue/437 search for profile by handle

ben-biddington:
Shows that a request to:

> /lb/v1/search_profiles?q=handle@authority

Invokes `ISearchProvider.SearchProfiles` and requires authorization.

The fallback behaviour can be shown elsewhere.

> Lookup profiles by handle. Check the local data set and fallback to webfinger if we don't have it.

## How it works

* [Network requests to `/lb/v1/search_profiles?q={handle}` invoke ISearchProvider.SearchProfiles](https://github.com/ben-biddington/Letterbook/blob/c9bb57fdbeac714393aa8ba2c81c4f4f88e8feac/Tests/Letterbook.IntegrationTests/ProfileLookupTests.cs)
* [`FallbackSearchProvider` is the implementation of `ISearchProvider`](https://github.com/ben-biddington/Letterbook/blob/42858f9eba0e6f6e74a89b0507eaaf9b229baf24/Tests/Letterbook.IntegrationTests/ProfileLookupTests.cs#L65) with primary `DataAdapter` and secondary `WebFingerClient`.
* [`FallbackSearchProvider` falls back to secondary provider when primary returns no result](https://github.com/ben-biddington/Letterbook/blob/c9bb57fdbeac714393aa8ba2c81c4f4f88e8feac/Tests/Letterbook.Adapter.ActivityPub.Test/FallbackSearchProviderTests.cs)

Happy to move things around if you don't like where they are now.

## Questions

* ~~Should "no results found" return 404 instead?~~
  [edit] Thinking about this, it would be confusing because the resource *is* there: the search algorithm is the resource
* Should an unauthorized request return 401 instead of 302 to `/Identity/Account/Login``?

## Related
- https://github.com/Letterbook/Letterbook/issues/437


Tests/Letterbook.IntegrationTests/AccountsApiTests.cs
```diff
@@ -65,24 +64,15 @@ dotnet user-secrets clear --project Source/Letterbook
...
-		await AddInviteCode("abc");
-
 		var response = await _client.PostAsync("/lb/v1/user_account/register", JsonContent.Create(new RegistrationRequest
 		{
 			Handle = "example-handle",
-			InviteCode = "abc",
+			InviteCode = _inviteCode.Code,
 			Password = "$Password1",
 			ConfirmPassword = "$Password1",
 			Email = "anything@domain.com",
```
jenniferplusplus: This is a small thing, but I would prefer if the URLs we use in tests and examples cannot resolve to anything, unless it's a domain we control.

So, generally, tests should use the .example tld. In some cases it may also make sense to use example.com, example.org, letterbook.com, or letterbookhq.com.

ben-biddington: Good call

ben-biddington:
I think we're getting pretty close to ready now, @jenniferplusplus -- just had to [add the ignore in one more place](https://github.com/Letterbook/Letterbook/pull/482/changes/f3901a5fd094d61ec361643ae49f90723e63f217) to stabilise the build. Lettuce know what you think

